PR #46, opened 2014-12-16
design tweaks to the stream title to make align nicer

pauloneillft:
@commuterjoy please have a look I only changed one very small thing. Should be ok to merge :)


wheresrhys:
Looks like a markup change has made this redundant

PR #56, opened 2015-01-02
Adding flag check

aintgoin2goa:


server/view-helpers/if-is-ad-slot.js
```diff
@@ -1,5 +1,10 @@
 "use strict";
 module.exports = function(index, options) {
+	var adsFlag = options.data.root.flags.ads;
+	if(!adsFlag || adsFlag.isSwitchedOff){
```
matthew-andrews: Can we change this to:-
```js
var adsFlag = options.data.root.flags.ads;
if (!adsFlag || !adsFlag.isSwitchedOn){
```

I want to deprecate [`isSwitchedOff`](https://github.com/Financial-Times/next-feature-flags-client/issues/26)

matthew-andrews: actually never mind

matthew-andrews: we'll frist decide whether or not to deprecate it then remove it we do that

PR #59, opened 2015-01-06
Adding grunt task for sourcemaps

aintgoin2goa:
The task extracts the sourcemap data from main.js transforms the sources to urls rather than local files and spits it out into a seperate .map file.

This should help for debugging production errors and (if sentry supports sourcemaps) should lead to better errors.

It can be split out into it's own module once we've verified it works as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ run-router:
 	export grumman=${PORT}; export PORT=5050; export DEBUG=proxy ; next-router
 
 build:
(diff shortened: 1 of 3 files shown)


aintgoin2goa:
Anyone have an opinion on this?  Can i go ahead and merge it and see if it works?

aintgoin2goa:
Ok, this doesn't actually work - will close for now

PR #63, opened 2015-01-08
Sourcemap grunt task

aintgoin2goa:
Ok, it works this time - it is a riduculous solution, but it works.

Bascially, we need run origami build tools, then do the minification ourselved, using the grunt-sourcemaps plugin to create the new sourcemap.

Unfortunately, grunt-sourcemaps does some weird stuff with paths so we save the new sourcemap inline, then extract it back out, transform all the paths to match the app and chuck it in a separate file.

Once this is merged I can verify it works in production and that sentry can use it, before attempting to either apply this across all the apps or get it into origami-build-tools

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -58,22 +58,20 @@ gulp.task('build', function () {
 		sass: './client/main.scss',
 		js: './client/main.js',
 		buildFolder: './public',
-		env: process.env.ENVIRONMENT || 'production',
-		sourcemaps : true,
-		compress : {js:false, scss:true}
+		env: 'development'
 	});
 });
 
-gulp.task('minify-js',['build', 'sourcemap'], function(){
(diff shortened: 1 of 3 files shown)


aintgoin2goa:
OK, this is a better version - but the only way to test it is to get it in production, as sentry processes the sourcemap at their end (I think).

gulpfile.js
```diff
@@ -3,19 +3,62 @@
...
+
 
 gulp.task('build', function () {
-	obt.build(gulp, {
+	return obt.build(gulp, {
 		sass: './client/main.scss',
 		js: './client/main.js',
 		buildFolder: './public',
-		env: process.env.ENVIRONMENT || 'production'
+		env: 'development'
```
matthew-andrews: i think this is going to not compress sass?

PR #77, opened 2015-01-14
Videos render in article page

leanne1:


client/components/video/main.js
```diff
@@ -0,0 +1,88 @@
...
+
+var video = {
+	'video.ft.com': [],
+	'youtube.com': [],
+	'player.vimeo.com': []
+};
+
+var videoTags = [].slice.call(document.querySelectorAll('a'));
+
+(function initVideos () {
```
matthew-andrews: if you're self executing then you don't need to wrap it in a function at all

matthew-andrews:
otherwise :+1:

matthew-andrews:
:+1:

PR #80, opened 2015-01-15
Data table styling

leanne1:


client/main.scss
```diff
@@ -38,6 +38,7 @@ body {
 @import "next-ads-component/main";
 $o-quote-is-silent: false;
 @import "o-quote/main";
+@import "o-table/main";
```
matthew-andrews: do we need `$o-table-is-silent: false` ?

PR #82, opened 2015-01-15
Make sure that you have some ids from the query 

aintgoin2goa:
Fix for this bug: https://app.getsentry.com/nextftcom/ft-next-grumman/group/45220490/

Grumman was not finding any ids for this primary theme and passing and empty array back to the api client, which then errored.


matthew-andrews:
:+1:

PR #84, opened 2015-01-15
fixing potential race condition

aintgoin2goa:
Realised that this still a potential for a race condition, now checks if the polyfills event has already been fired before listening for it


matthew-andrews:
:+1:

PR #99, opened 2015-02-06
Typography changes targeted to BP small

leanne1:


.gitignore
```diff
@@ -15,3 +15,4 @@ npm-debug.log
 .editorconfig
 public/main.css-asset-hash.json
 public/main.js-asset-hash.json
+public/main.js.map
```
matthew-andrews: do we have js sourcemaps now?  very cool if so…

client/main.scss
```diff
@@ -1,4 +1,5 @@
 $o-grid-mode: snappy;
+$o-grid-debug-mode: true;
```
matthew-andrews: do we want this in prod?

client/main.scss
```diff
@@ -191,6 +189,26 @@ $inlineImgWidth: 300px;
...
+	}
+	.article__body {
+		font-size: 21px;
+		p {
+			line-height: 28px;
+		}
+	}
+	.article__img-inline,
+	.article__img-inline-offset {
+		max-width: $inlineImgWidth;
```
matthew-andrews: i think we usually have `$inline-img-width` style capitalisation for sass vars, but not a big deal

PR #105, opened 2015-02-09
Ensure ads flag is respected

aintgoin2goa:


client/components/more-on/main.js
```diff
@@ -16,15 +17,16 @@ Promise.allSettled = function(promises){
 	return Promise.all(promises.map(resolveWhenSettled));
 };
 
-var initAds = (function(){
+var initAds = function(flags){
 	var called = false;
 	return function(){
-		if(!called){
+		if(!called && flags.ads.isSwitchedOn){
+			var ads = require('next-ads-component');
```
matthew-andrews: we really ought to make our code such that we can `require` modules without them running code…

matthew-andrews: inline `require` statements will cause us problems if/when we want to change over to different systems & increase the danger of circular dependencies

aintgoin2goa: our code doesn't run - the ads component has an init method.  The problem is o-ads which has a couple of require('module')() bits - which is the cause of the error we're seeing.

So this is the only way to ensure that the code that causes the current error message never runs.

PR #106, opened 2015-02-10
Better author linking

matthew-andrews:
cc @leanne1


client/components/ad-slots/main.js
```diff
@@ -12,13 +12,13 @@ function placeMpu () {
 	var breakpoint = getCurrentBreakpoint().toLowerCase();
 	if (breakpoint === 'l' || breakpoint === 'xl') {
 		mpuSlotL.appendChild(dummyAd);
-		mpuSlotXs.innerHTML('');
+		mpuSlotXs.innerHTML = '';
 	} else {
 		mpuSlotXs.appendChild(dummyAd);
-		mpuSlotL.innerHTML('');
+		mpuSlotL.innerHTML = '';
```
matthew-andrews: i think this was you :smile:

PR #111, opened 2015-02-10
Rwd img

leanne1:


client/main.scss
```diff
@@ -65,6 +65,10 @@ $inline-img-width: 300px;
 .article__color-brand {
 	padding-bottom: $brandOffset;
 
+	// &-offset {
+	// 	background-color: #f6e9d8;
+	// }
```
matthew-andrews: would you mind deleting this? if we need it back we can always use git history :)

views/layout_2-improved.html
```diff
@@ -1,6 +1,7 @@
 {{#if article.id}}
 <article class="article article--default" data-capi-id="{{article.id}}">
 	<div class="article__color-brand">
+	<!-- {{#if article.byline}} article__color-brand{-offset{{/if}} -->
```
matthew-andrews: same as above re: commented out code

matthew-andrews:
this is really cool—responsive images!!!

PR #112, opened 2015-02-10
No-image layout hacks

leanne1:


client/main.scss
```diff
@@ -1,5 +1,5 @@
 $o-grid-mode: snappy;
-//$o-grid-debug-mode: true;
+$o-grid-debug-mode: true;
```
matthew-andrews: did we need this?

views/layout_2-improved.html
```diff
@@ -31,9 +31,16 @@ <h3 class="article__title">{{{article.titleHTML}}}</h3>
 					</div>
 				</div>
 			</div>
-			<div data-o-grid-colspan="12 L8 XL7">
-				<img src="{{#resize 700}}http://ft-next-grumman.herokuapp.com/{{article.id}}/main-image{{/resize}}" class="article__img-main" onerror="this.parentNode.style.display='none';"/>
+			<script>
+			function noImgHack () {
+				document.querySelector('.no-img-img').style.display = 'none';			
+				document.querySelector('.article').classList.add('no-img-hack');
```
matthew-andrews: i'm afraid we can't rely on `querySelector` or `classList` because they would depend on the polyfill service

views/layout_2-improved.html
```diff
@@ -31,9 +31,16 @@ <h3 class="article__title">{{{article.titleHTML}}}</h3>
...
-			<div data-o-grid-colspan="12 L8 XL7">
-				<img src="{{#resize 700}}http://ft-next-grumman.herokuapp.com/{{article.id}}/main-image{{/resize}}" class="article__img-main" onerror="this.parentNode.style.display='none';"/>
+			<script>
+			function noImgHack () {
+				document.querySelector('.no-img-img').style.display = 'none';			
+				document.querySelector('.article').classList.add('no-img-hack');		
+			}
+			</script>
+			<div data-o-grid-colspan="12 L8 XL7" class="no-img-img">
+				<img src="{{#resize 700}}http://ft-next-grumman.herokuapp.com/{{article.id}}/main-image{{/resize}}" class="article__img-main" onerror="noImgHack()"/>
```
matthew-andrews: `onerror="this.parentNode.className='no-image'; this.parentNode.innerHTML='&nbsp;'"` ?

views/layout_2-improved.html
leanne1: querySelector will work on IE8+ against CSS 2 selectors, so this should be ok?

matthew-andrews: chatted with @leanne1 and we decided as this was a hack we'll probably get away with it…

views/layout_2-improved.html
leanne1: The class is being added to the .article element higher up the DOM tree, rather than the parent of the image, as it uses the css cascade to apply styles to multiple elements

matthew-andrews: yep fair enoug

PR #118, opened 2015-02-11
avoid new header

wheresrhys:


matthew-andrews:
wai?

matthew-andrews:
[new is always better](https://www.youtube.com/watch?v=x7_KFazcb3w)

wheresrhys:
nah - it breaks - will figure out why tomorrow

PR #130, opened 2015-02-19
Nightwatch tests

spriggsha:
Adding basic nightwatch framework for grumman, plus the provisioning steps in the makefile to generate heroku branches.


commuterjoy:
:+1: lets get this merged, the tests fail for me, so I've disabled them, but we can fix tomorrow am.

PR #140, opened 2015-02-25
Provide get out of jail free card

matthew-andrews:


matthew-andrews:
// cc @aintgoin2goa

server/controllers/capi.js
```diff
@@ -118,7 +118,18 @@ module.exports = function(req, res, next) {
 		})
 		.catch(function(err) {
 			if (err instanceof fetchres.BadServerResponseError) {
-				res.status(404).end();
+				fetch('http://api.ft.com/content/items/v1/' + req.params[0], {
```
commuterjoy: Can we do this from the next-errors app?

matthew-andrews: It's quite specific to articles at the moment.

PR #141, opened 2015-02-25
Remove ads

matthew-andrews:


matthew-andrews:
@robin-marr-ft sorry i had to remove all your advert code — you really should have talked to George Bogdis (his designs have slots for adverts in) — can't just shove adverts inside the body content becasue it breaks the editorial components

PR #145, opened 2015-02-26
prepare for images

matthew-andrews:


commuterjoy:
:+1:

server/transforms/ft-content.js
```diff
@@ -6,6 +6,15 @@ module.exports = function(index, el) {
...
+	var type = el.attr('type');
 	url = url.replace('http://api.ft.com/content/', '/');
-	return '<a href="' + url + '">' + text + '</a>';
+
+	switch (type) {
+		case 'http://www.ft.com/ontology/content/ImageSet':
+			return '<img src="/embedded-components/image' + url + '"/ >';
+		case 'http://www.ft.com/ontology/content/Article':
+			return '<a href="' + url + '">' + text + '</a>';
+		default:
```
commuterjoy: i think this is nice - we can work at a slightly different pace than the api team and decouple when it appears in the API from when we show it to users

PR #153, opened 2015-02-27
Ads v3

aintgoin2goa:


views/partials/articles/capi.html
```diff
@@ -101,9 +101,6 @@ <h3 class="next-card--article__headline next-card--article__headline--large" dat
 
     <div class="next-card--article__body">
         {{{paragraphs body start=0 end=3}}}
-        {{#if @root.flags.ads.isSwitchedOn}}
```
matthew-andrews: eek i think these templates are no more :)

matthew-andrews:
:+1: don't worry about the comment — i've made a note fo rme to tidy this up

PR #165, opened 2015-03-03
Pull out image & quote components on desktop views

matthew-andrews:


matthew-andrews:
cc @ironsidevsquincy

ironsidevsquincy:
:+1:

PR #168, opened 2015-03-03
Swtich to metric for article fonts

matthew-andrews:


matthew-andrews:
cc @ironsidevsquincy

ironsidevsquincy:
:+1:

PR #171, opened 2015-03-03
Timeout fetch after 3000ms

### LINKED ISSUE #170: add timeout
although fetch deosnt' support this, the server side implementation of fetch that we're doing does…

matthew-andrews:
fixes #170


ironsidevsquincy:
could create a wrapper round `fetch` that sets up these config opts (timeout, headers) as defaults

:+1:

commuterjoy:
:+1: grand, thanks

PR #183, opened 2015-03-05
switch access to middleware

aintgoin2goa:
It turns out that Fastly turns a HEAD request into a GET so switching to middleware and pivoting on the header field instead.


matthew-andrews:
> Fastly turns a HEAD request into a GET

whatttttt

( :+1: )

PR #185, opened 2015-03-05
Sourcemaps in production

aintgoin2goa:
Sadly uglify-js fails to carry the sourceContent across properly but this will be fixed later.

At least errors should now be attributed to the file they actually came from...


matthew-andrews:
weird — origami build tools installed fine all day and now this?

PR #187, opened 2015-03-06
Sanity check video id

ironsidevsquincy:
Before making request to brightcove

https://app.getsentry.com/nextftcom/ft-next-frontend/group/56963178/

### DIFF
--- a/client/components/video/brightcove.js
+++ b/client/components/video/brightcove.js
@@ -2,6 +2,11 @@
 'use strict';
 
 function brightcove(videoId) {
+
+	if (!videoId.match(/^\d{13}$/)) {
+		return Promise.reject(new Error('Invalid video id: ' + videoId));
+	}
+
 	return fetch('http://ft-next-brightcove-proxy-api.herokuapp.com/' + videoId)
 		.then(function(response) {
 			if (response.status === 404) {


commuterjoy:
it's worth reporting this as an upstream problem isn't it? where are the bad video id's coming from?

ironsidevsquincy:
Reported and fixed. I think this is still a useful sanity check to have

matthew-andrews:
my preference would be not to fix it in grumman if it's already fixed upstream

matthew-andrews:
oh wait this just throws an error, that's fine.  will make it easier to debug if reoccurs…

client/components/video/brightcove.js
```diff
@@ -2,6 +2,11 @@
 'use strict';
 
 function brightcove(videoId) {
+
+	if (!videoId.match(/^\d{13}$/)) {
+		return Promise.reject(new Error('Invalid video id: ' + videoId));
```
matthew-andrews: any reason why you do this an not
```js
throw new Error('Invalid video id: ' + videoId);
```

ironsidevsquincy: pretty sure it didn't work for some reason, got the `Uncaught (in promise)...` error in the console

matthew-andrews: weird, that should work.  would you mind showing me?

ironsidevsquincy: Rather, https://github.com/Financial-Times/next-grumman/blob/master/client/components/video/main.js#L27 fails when calling `then` on a returned `Error` object

matthew-andrews: ah, it shouldn't be `return new Error`, should be just `throw new Error(…);`

PR #193, opened 2015-03-10
Article header design update

ironsidevsquincy:
header


commuterjoy:
nice design work :)

matthew-andrews:
:+1:

PR #195, opened 2015-03-10
Fail faster

matthew-andrews:


matthew-andrews:
cc @ironsidevsquincy

matthew-andrews:
screen shot 2015-03-10 at 10 49 15
to avoid the not very good “timeout after 2000ms” message. 
http://mochajs.org/

ironsidevsquincy:
:+1:

PR #196, opened 2015-03-10
Notify when assets built

ironsidevsquincy:


matthew-andrews:
what was this for?

ironsidevsquincy:
saves watching terminal for build to end

matthew-andrews:
:+1:

PR #197, opened 2015-03-10
Some fixes, some removing hacks, some adding hacks to make master images work somewhat more

matthew-andrews:


client/components/article/main.scss
```diff
@@ -166,30 +166,6 @@ $inline-img-width: 300px;
 		width: 100%;
 	}
 }
-.article__main-image {
-	position: relative;
-	margin-top: -$brand-offset;
-	padding-bottom: 9% * 100% / 16%; // 56.25%;
```
matthew-andrews: i'll re-add this logic in #198

client/components/article/main.scss
```diff
@@ -166,30 +166,6 @@ $inline-img-width: 300px;
 		width: 100%;
 	}
 }
-.article__main-image {
-	position: relative;
-	margin-top: -$brand-offset;
```
ironsidevsquincy: these was to overlap the image over the header slightly

ironsidevsquincy:
:+1:

client/components/article/main.scss
matthew-andrews: i'll add that back

PR #200, opened 2015-03-10
Revert "Mark external inline images as 'article__inline-image's"

ironsidevsquincy:
Reverts Financial-Times/next-grumman#199

Causing this - 

screen shot 2015-03-10 at 15 01 01


matthew-andrews:
oops sorry

PR #201, opened 2015-03-10
udated ads and header

wheresrhys:
@matthew-andrews @ironsidevsquincy will mean user prefs buttons are properly controlled by flags


ironsidevsquincy:
:+1:

PR #205, opened 2015-03-11
Fix image watercolour (which I broke)

matthew-andrews:


matthew-andrews:
@ironsidevsquincy

matthew-andrews:
See the api response here: http://api.ft.com/content/08e9e07a-c746-11e4-8e1f-00144feab7de

views/layout.html
```diff
@@ -1,6 +1,7 @@
 <article class="article article--default{{#if showTOC}} article--toc{{/if}}" data-capi-id="{{article.id}}">
 	<div class="article__header">
-		<div class="article__header__bg" style="background-image: url({{#resize 1400}}http://ft-next-grumman-v002.herokuapp.com/{{article.id}}/main-image{{/resize}})"></div>
+		{{! TODO: Re-add resize helper (1400) once CAPI Content-Type bug fixed }}
```
ironsidevsquincy: also need to look into making it responsive - could even load in a smaller version and stretch it, as we now slightly blur the image

ironsidevsquincy:
:+1:

PR #206, opened 2015-03-11
Centralise capi v1

matthew-andrews:
- Fix many blogs (which previously wouldn't have worked on article endpoints)
- Enables usage of elastic search in more places
- Slightly tidier code


server/controllers/access.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 module.exports = function(req, res, next) {
-	var apiKey = res.locals.flags.articlesFromContentApiV2.isSwitchedOn ? process.env.apikey : process.env.api2key;
+	var apiKey = process.env.apikey;
```
matthew-andrews: @aintgoin2goa I think the ft-api-client only supports api v1 currently I would have thought this would have broken things.

ironsidevsquincy:
:+1:  - smoke tests?

matthew-andrews:
i know…  `nock` is doing something i don't understand

PR #208, opened 2015-03-11
Update Table of Contens to use o-forms v1

ironsidevsquincy:


matthew-andrews:
:+1:

matthew-andrews:
is select the best UI here?

maybe one for george b but this sort of thing might be easier for users, maybe:
http://financial-times.github.io/next/docs/overview/principles/

ironsidevsquincy:
that would be the ideal, the issue we had was were to put it. once the current header rejig settles down, we might find a space for it

PR #212, opened 2015-03-11
Inline elements tidy

ironsidevsquincy:
use latest origami pull quotes and big number, tidy up/remove unused classes

also, now have a `ng-pull-out` for pulled-out elements, and `ng-inline-element` for, er, inline elements

screen shot 2015-03-11 at 16 09 20

### DIFF
--- a/client/_helpers.scss
+++ b/client/_helpers.scss
@@ -3,7 +3,7 @@
 	display: block;
 	margin: 35px auto;
 
-	@include oGridRespondTo(M) {
+	@include oGridRespondTo(S) {
 		@include box-sizing(border-box);
 		float: left;
 		clear: left;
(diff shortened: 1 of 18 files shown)


client/_helpers.scss
```diff
@@ -0,0 +1,18 @@
+// GRUMMAN HELPER CLASSES
+.ng-inline-element {
```
matthew-andrews: ng?

matthew-andrews:
:+1:

client/_helpers.scss
ironsidevsquincy: next-grumman

client/components/article/main.scss
```diff
@@ -191,26 +153,19 @@ $inline-img-width: 300px;
 	text-align: center;
 	padding: 10px 0;
 }
-.article__img-inline,
-.article__img-inline-offset {
```
ironsidevsquincy: couldn't find references to these anywhere, assuming they're an antiquated name for `article__inline-image`

PR #219, opened 2015-03-12
Handle different response from elasticsearch

ironsidevsquincy:


server/controllers/capi.js
```diff
@@ -105,7 +105,7 @@ module.exports = function(req, res, next) {
 
 					res.render('layout', {
 						article: article,
-						articleV1: articleV1.item,
+						articleV1: res.locals.flags.elasticSearchItemGet.isSwitchedOn ? articleV1._source.item : articleV1.item,
```
matthew-andrews: hmm…  this probably needs fixing everywehre we use `fetch-capi-v1` :(

PR #220, opened 2015-03-12
Track article save button

matthew-andrews:


matthew-andrews:
is this right @adambraimbridge ?

commuterjoy:
:+1: 

> is this right

yes, `save-for-later` is the convention.

matthew-andrews:
sweets

PR #223, opened 2015-03-12
Inline related component

ironsidevsquincy:
screen shot 2015-03-12 at 14 56 05

### DIFF
--- a/server/controllers/capi.js
+++ b/server/controllers/capi.js
@@ -55,7 +55,7 @@ module.exports = function(req, res, next) {
 	Promise.all([articleV1Promise, articleV2Promise])
 		.then(fetchres.json)
 		.then(function(articles) {
-			var articleV1 = articles[0],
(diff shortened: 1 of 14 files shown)


matthew-andrews:
i've fixed the build failure ( i think ) so hold off on merging this and once that's merged we can rebase/merge from that into this then merge this?

server/controllers/capi.js
```diff
@@ -78,6 +78,12 @@ module.exports = function(req, res, next) {
 					$('blockquote').attr('class', 'article__block-quote o-quote o-quote--standard');
 					$('pull-quote').replaceWith(pullQuotesTransform);
 
+					// insert test related
+					if (articleV1._source.item.package.length > 0 && $('ft-paragraph').length >= 6) {
```
matthew-andrews: hmm…  i thought `_source` was elastic search only?

matthew-andrews: we should support elastic search on and off (try toggling the flag)

ironsidevsquincy: absolutely, silly mistake. another reason to move into the `fetchCapiV1`

PR #227, opened 2015-03-12
Trim whitespace around links

matthew-andrews:


matthew-andrews:
… and kill old authors endpoint

PR #228, opened 2015-03-12
Not in api v2

matthew-andrews:


server/controllers/capi.js
```diff
@@ -76,7 +76,7 @@ module.exports = function(req, res, next) {
 					$('pull-quote').replaceWith(pullQuotesTransform);
 
 					// insert test related
-					if (articleV1.item.package.length > 0 && $('ft-paragraph').length >= 6) {
```
matthew-andrews: @ironsidevsquincy this isn't scheduled for the new api any time soon, and seems like an optimisation rather than essential…

ironsidevsquincy: quite right

ironsidevsquincy:
:+1:

PR #230, opened 2015-03-13
Use correct background image for inline related

ironsidevsquincy:


matthew-andrews:
oh yeah i noticed this

matthew-andrews:
:+1:

PR #234, opened 2015-03-13
Tracking smoke test

ironsidevsquincy:
Check that all links have a `data-trackable` attribute

Also, nook was desperate for a response from `http://api.ft.com/site/v1/pages?apiKey=12345`


matthew-andrews:
fail build if we forget tracking… :+1: nice!

tests/server/smoke-spec.js
```diff
@@ -140,4 +144,26 @@ describe('smoke tests for the app', function () {
 		});
 
 	});
+
+	describe('tracking', function () {
```
commuterjoy: nice. need to apply this everywhere.

PR #235, opened 2015-03-13
No more fancy, bandwidth-hogging backgrounds

ironsidevsquincy:
And main image is now inlined

screen shot 2015-03-13 at 16 36 20


matthew-andrews:
main image inlined on the LHS may have been veto'd by stakeholders — one to check with Dan S / George B — otherwise :+1:

commuterjoy:
Personally I like the main image on the left

1. Often the FT's images are quite dull.  
2. It makes the design feel less like a medium et al rip off.

commuterjoy:
3. Makes the text closer to the top off the page, which is more useful to the user.

matthew-andrews:
let's see what they say next week

PR #240, opened 2015-03-16
Starting to work on improved related content

matthew-andrews:
not for merging yet but just wanted to give a quick preview of the direction i'm thinking about going @ironsidevsquincy


server/controllers/more-on.js
```diff
@@ -1,42 +1,59 @@
...
-							.forEach(function(item) {
-								stream.push('methode', item);
-							});
-						res.set(cacheControl);
-						res.render('more-on' + (req.query.view ? '-' + req.query.view : ''), {
-							mode: 'expand',
-							stream: stream.texturedItems,
-							title: 'See also'
+			return Promise.all(moreOns.map(function(moreOn) {
+					return fetchCapiV2({ uuid: moreOn.id })
```
matthew-andrews: using capi v2 helps us promote fewer articles that won't work on Next

server/controllers/access.js
```diff
@@ -1,16 +1,36 @@
...
-			res.set('X-FT-Content-Classification', article.contentClassification);
-			res.status(200).end();
-		}).catch(next);
+		fetchCapiV1({
+			uuid: req.params[0],
+			useElasticSearch: res.locals.flags.elasticSearchItemGet.isSwitchedOn
+		})
+			.then(function(article) {
+				var results = /cms\/s\/([0-3])\//i.exec(article.item.location.uri);
+				var classification = 'unconditional';
```
matthew-andrews: https://github.com/Financial-Times/ft-api-client/blob/5e1488b80c3d6c3cdfdb872f91f2f498a9de61f9/lib/models/article.js#L502-525

ironsidevsquincy:
:+1:

PR #243, opened 2015-03-16
Fix subhead ids

ironsidevsquincy:
Just use a standard, incrementing id, rather than fancily trying to convert the text into an id


matthew-andrews:
nice, simple :+1:

PR #250, opened 2015-03-17
Relative dates not appearing in moreons

matthew-andrews:
@ironsidevsquincy

### DIFF
--- a/client/_mixins.scss
+++ b/client/_mixins.scss
@@ -1,12 +1,6 @@
 @mixin filter($funcs...) {
-	-webkit-filter: $funcs;
-	-moz-filter: $funcs;
-	-o-filter: $funcs;
-	-ms-filter: $funcs;
 	filter: $funcs;
 }
 @mixin box-sizing($value) {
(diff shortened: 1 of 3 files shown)


client/_mixins.scss
```diff
@@ -1,12 +1,6 @@
 @mixin filter($funcs...) {
-	-webkit-filter: $funcs;
-	-moz-filter: $funcs;
-	-o-filter: $funcs;
-	-ms-filter: $funcs;
```
ironsidevsquincy: definitely not needed? e.g latest chrome only works with `webkit` prefix

matthew-andrews: origami-build-tools ships with autoprefixer

matthew-andrews: > autoprefixerBrowsers: Array An array of strings of browser names for autoprefixer to check what prefixes it needs. (Default: ["> 1%", "last 2 versions", "ie > 6", "ff ESR"])

ironsidevsquincy: a-mazing

PR #262, opened 2015-03-18
Correct asset building when in prod

ironsidevsquincy:
Building JS was failing if `ENVIRONMENT` was prod, due to obt stripping out the sourcemap comment

Also, no need to minify/create external source map if in dev


matthew-andrews:
lgtm :+1:

PR #263, opened 2015-03-18
Better debug when grumman goes down

matthew-andrews:
cc @ironsidevsquincy 

and fix a bug where 404s wouldn't render in most cases.


server/controllers/capi.js
```diff
@@ -146,7 +146,7 @@ module.exports = function(req, res, next) {
 		})
 		.catch(function(err) {
 			if (err instanceof fetchres.BadServerResponseError) {
-				fetchres(req.params[0])
+				fetchCapiV1(req.params[0])
```
matthew-andrews: oops

ironsidevsquincy:
:+1:

PR #272, opened 2015-03-19
Don't promote 404 errors to sentry

matthew-andrews:


matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-grumman/group/56122933/

PR #278, opened 2015-03-19
Styling of 'more on'

ironsidevsquincy:
* Consistent naming - 'more on' everywhere, not 'related'
 * `/more-on/:id` returns article's story package, `/more-on/:metadata/:id` articles associated with that topic

screen shot 2015-03-19 at 16 56 54


matthew-andrews:
lgtm :+1:

PR #299, opened 2015-03-23
Add empty alt attribute

kaelig:
An empty alt attribute is preferable to no attribute for accessibility.

Otherwise screen readers read the `src` out loud.


kaelig:
Actually forget it, I'll just make tests fail and then you can fix stuff :)

PR #300, opened 2015-03-23
Images should have an alt attribute

kaelig:
An empty alt attribute is preferable to no attribute for accessibility.

Otherwise screen readers read the `src` out loud.


kaelig:
This PR is only here to make tests fail. Feel free to close it if it's already on your radar.

PR #330, opened 2015-03-30
Get V2 metadata and /content api requests are (currently) always articles

matthew-andrews:
@ironsidevsquincy

### DIFF
--- a/server/utils/fetch-capi-v2.js
+++ b/server/utils/fetch-capi-v2.js
@@ -6,8 +6,8 @@ var logger = require('./logger');
 
 module.exports = function(opts) {
 	var uuid = opts.uuid;
+	var type = opts.type || 'unknown';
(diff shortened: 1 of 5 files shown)


server/utils/fetch-capi-v2.js
```diff
@@ -20,7 +19,7 @@ module.exports = function(opts) {
 				if (!response.ok) {
 					logger.warn('Failed getting CAPIv2 content', {
 						uuid: uuid,
-						type: type,
+						type: 'Article',
```
matthew-andrews: @ironsidevsquincy is there something I've missed?  I think this is (currently) always articles

matthew-andrews:
Gives you V2 metadata:-
screen shot 2015-03-30 at 15 19 21

server/utils/fetch-capi-v2.js
ironsidevsquincy: Image sets - https://github.com/Financial-Times/next-grumman/blob/master/server/transforms/images.js#L12

Can now see in the log when images are failing

matthew-andrews: ohhh

PR #336, opened 2015-03-31
Don't run JS on error pages

ironsidevsquincy:


bower.json
```diff
@@ -3,16 +3,16 @@
   "dependencies": {
     "o-fonts": "^1.6.4",
     "next-feature-flags-client": "^v5.1.0",
-    "next-ads-component": "https://github.com/Financial-Times/next-ads-component.git#^4.0.0",
```
matthew-andrews: much neater :+1:

PR #338, opened 2015-03-31
Better <p><strong> to <h3> transform

ironsidevsquincy:
Only transform if the `p` is at the top level (e.g. stop catching `<p><strong>`s in block quotes)


matthew-andrews:
this is soon going to be unnecessary :-)

matthew-andrews:
:+1:

PR #342, opened 2015-03-31
Broken FastFT posts

matthew-andrews:


server/controllers/capi.js
```diff
@@ -121,7 +121,7 @@ module.exports = function(req, res, next) {
 						});
 
 					// Update the images (resize, add image captions, etc)
-					images($body, res.locals.flags)
+					return images($body, res.locals.flags)
```
matthew-andrews: @ironsidevsquincy without this the promise failures don't get pushed up to sentry :-)

PR #347, opened 2015-04-01
Move slideshows outside of containing <p>

ironsidevsquincy:


matthew-andrews:
i think this should be fixed upstream…  can we add a note to unwind this in the future?

ironsidevsquincy:
I believe they're working on this - someone sent a message on slack, which I'm struggling to find.

This logic wont affect anything when it's fixed, but I'll add a comment to remove

PR #352, opened 2015-04-07
Test with elastic search turned off

ironsidevsquincy:
Simple test when `elasticSearchItemGet: off`

Also, remove unused fixtures, and add `update-flags` make task


ironsidevsquincy:
addresses https://github.com/Financial-Times/next-grumman/issues/271

PR #358, opened 2015-04-08
Limit width of inline image's caption

ironsidevsquincy:
### Before

screen shot 2015-04-08 at 14 27 00

### After

screen shot 2015-04-08 at 14 26 42


matthew-andrews:
:+

PR #360, opened 2015-04-09
FastFT bug + failing test

matthew-andrews:


matthew-andrews:
@ironsidevsquincy

server/controllers/capi.js
```diff
@@ -130,7 +138,7 @@ module.exports = function(req, res, next) {
 									};
 								}).get(),
 								showTOC: res.locals.flags.articleTOC.isSwitchedOn && $subheaders.length > 2,
-								isColumnist: articleV1.item.metadata.primarySection.term.name === 'Columnists',
+								isColumnist: isColumnist,
```
ironsidevsquincy: naively assumed there would always be a primarySection...

ironsidevsquincy: actually that might be the case, issue here was no `articleV1` - probably better to safe though

matthew-andrews: ah.  that is probably a correct assumption.  The false assumption was that the content would always be in CAPI v1 ;-)

http://api.ft.com/content/items/v1/b002e5ee-3096-3f51-9925-32b157740c98?apiKey=XXX => 404
http://api.ft.com/content/b002e5ee-3096-3f51-9925-32b157740c98?apiKey=XXX => 200

matthew-andrews: could probably simplify the check to just be `if (articleV1)` to be honest… would you like me to do this?

ironsidevsquincy: if that assumption (always a primarySection) correct, than yeah, might as well simplify (I might do the same to https://github.com/Financial-Times/next-grumman/blob/master/server/transforms/byline.js#L4)

PR #361, opened 2015-04-09
Tidy directory structure of tests

ironsidevsquincy:
Mainly just moving files about

Also split the smoke tests into separate files


matthew-andrews:
seems good

PR #363, opened 2015-04-09
Display people/orgs even if no v1-to-v2-mapping

ironsidevsquincy:
As currently we're not using v2 data


matthew-andrews:
i'm using v2 data behind “mentionsV2” flag…

ironsidevsquincy:
This is just for the people and organisations endpoints, that display the respective components - we're not displaying any v2 data in them, so using the v1-to-v2 mapping isn't absolutely necessary right now, but hopefully we'll start using it soon (e.g. linking to a stream page using the v2 uuid, etc)

matthew-andrews:
ah ok. sweet.

matthew-andrews:
:+1:

PR #371, opened 2015-04-10
Leave a space for the main image and video

ironsidevsquincy:
Cuts down on jumpy pages

### DIFF
--- a/client/_helpers.scss
+++ b/client/_helpers.scss
@@ -56,3 +56,12 @@
 	line-height: 1em;
 	opacity: 0.7;
 }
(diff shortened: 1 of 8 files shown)


matthew-andrews:
hmm slightly risky.  they can put any size inline image at the top of articles……  i think this is fine for a hack for now but we will be able to have better solution when we get size information through the API……

ironsidevsquincy:
Yeah, I did wonder that. Guessing video's always 16x9(?).

Ok, will release for now so things are a bit smoother, but will revisit once we get size data

PR #375, opened 2015-04-13
Annotations tidy up

ironsidevsquincy:
Single endpoint that returns people/organisation using either v1 tagging or v2 annotations depending on flag


matthew-andrews:
:+1:

PR #376, opened 2015-04-13
Remove FOIT

ironsidevsquincy:
*Flash Of Invisible Text

Occurs when downloading the font - use the Font event api to determine when the fonts have loaded before applying a class

Testing out here, can probably move into js-setup eventually

TODO - polyfill for older browsers


matthew-andrews:
Can go in next-js-setup?

ironsidevsquincy:
Yeah, moving it into js-setup

PR #377, opened 2015-04-13
Myft notifications

wheresrhys:
also removed a load of accidentally committed bower components and o-hoverable (as it's now provided by next-js-setup) @ironsidevsquincy @matthew-andrews


ironsidevsquincy:
:+1:, think those are the only places we're referencing `data-capi-id`

wheresrhys:
trying again with a cherrypick

PR #381, opened 2015-04-14
Display related tags on articles

ironsidevsquincy:
Use the same tag decision logic as used in streams

screen shot 2015-04-14 at 12 08 51

(Only display on L+)


server/controllers/capi.js
```diff
@@ -134,12 +134,31 @@ module.exports = function(req, res, next) {
...
-					var isColumnist;
 					// Some posts (e.g. FastFT are only available in CAPI v2)
-					if (articleV1) {
-						// TODO: Replace with something in CAPI v2
-						isColumnist = articleV1.item.metadata.primarySection.term.name === 'Columnists';
-					}
+					// TODO: Replace with something in CAPI v2
+					var isColumnist = articleV1 && articleV1.item.metadata.primarySection.term.name === 'Columnists';
+
+					var articleV1Metadata = articleV1 && articleV1.item.metadata;
```
matthew-andrews: sorry to be a pain but can we have a `mentionsV2` implementation too?  (Use V2 metadata and point semantic streams — organisations/uuid & people/uuid)

ironsidevsquincy: yep, of course

ironsidevsquincy:
@matthew-andrews, not showing first 5 annotations if `mentionsV2 ` is on (though this is pretty much what we get at the bottom of the article)

matthew-andrews:
cool

PR #384, opened 2015-04-14
using new notifications event

wheresrhys:
@ironsidevsquincy cherrypicked the one useful commit from previous attempt


ironsidevsquincy:
:+1:

PR #387, opened 2015-04-15
Take elastic search for another spin

matthew-andrews:


matthew-andrews:
The previous bug seems fixed!
http://ft-grumman-branch-7e6b105.herokuapp.com/3759c36a-e28c-11e4-aa1d-00144feab7de

PR #389, opened 2015-04-16
Tags should display correctly in a shallow header

ironsidevsquincy:
flex ftw

### Before

screen shot 2015-04-16 at 10 31 40

### After

screen shot 2015-04-16 at 10 31 43

resolves https://github.com/Financial-Times/next-grumman/issues/388

### DIFF
--- a/client/components/article/_main.scss
+++ b/client/components/article/_main.scss
@@ -14,7 +14,7 @@ $header-offset: 30px;
 	color: #ffffff;
 }
 .article__header-inner {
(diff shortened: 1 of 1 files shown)


matthew-andrews:
flextastic.  how does it look in old browsers?

ironsidevsquincy:
topics will be aligned to the top of the header rather than the bottom - acceptable, I think

matthew-andrews:
:+1:

PR #390, opened 2015-04-16
Beacon interactions with gallery

ironsidevsquincy:
Beacon when the gallery loads, and when the user interacts

### DIFF
--- a/client/components/slideshow/main.js
+++ b/client/components/slideshow/main.js
@@ -2,6 +2,7 @@
 
 var fetchres = require('fetchres');
 var Gallery = require('o-gallery');
+var Analytics = require('next-beacon-component');
 
 module.exports = function(els) {
 	[].slice.call(els).forEach(function(el) {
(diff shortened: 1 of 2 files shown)


client/components/slideshow/main.js
```diff
@@ -10,14 +11,27 @@ module.exports = function(els) {
...
 					container.innerHTML = data;
 					el.parentNode.replaceChild(container, el);
 					return container;
 				})
 				.then(function(el) {
+					var totalItems;
 					el.style.width = el.clientWidth + 'px';
+					el.addEventListener('oGallery.itemSelect', function (ev) {
+						if (ev.target.classList.contains('o-gallery--slideshow')) {
+							Analytics.fire('galleries', { item: ev.detail.itemID, totalItems: totalItems, interacted: true });
```
commuterjoy: Can this be `gallery` please?

commuterjoy: instead of item i'd be more specific - i.e. `picture`

matthew-andrews:
oh sorry……

client/components/slideshow/main.js
```diff
@@ -10,14 +11,27 @@ module.exports = function(els) {
...
+						}
+					});
+					el.addEventListener('oGallery.ready', function oGalleryIsReady(ev) {
+						if (ev.target.classList.contains('o-gallery--slideshow')) {
+							el.removeEventListener(ev.type, oGalleryIsReady);
+							totalItems = ev.target.querySelectorAll('.o-gallery__item').length;
+							Analytics.fire('galleries', { item: 0, totalItems: totalItems, interacted: false });
```
commuterjoy: what's the idea with this?

ironsidevsquincy: a way of knowing which pages have a gallery on them. maybe better to add it to the `page` data?

client/components/slideshow/main.js
ironsidevsquincy: I'll singularise™ and change to a more human word

client/components/slideshow/main.js
commuterjoy: Do this on the `next-beacon` project, on the server. I want to move away from using the client to track *everything* when it's more obvious to do on the server.

Each beacon call hits CAPI, so you could just add a `hasGallery` flag to the page.capi object

https://github.com/Financial-Times/next-beacon/blob/master/models/event.js#L84

PR #396, opened 2015-04-17
Basic v1 article

ironsidevsquincy:
For when the app fails to get v2 data

### DIFF
--- a/client/components/comments/main.js
+++ b/client/components/comments/main.js
@@ -5,7 +5,7 @@ var fetchres = require('fetchres');
 
 module.exports = {};
 module.exports.init = function(uuid, flags) {
-	if (flags.get('articleComments').isSwitchedOff || !document.querySelector('#comments')) {
+	if (!flags.get('articleComments').isSwitchedOn || !document.querySelector('#comments')) {
 		return;
 	}
 	return fetch('/' + uuid + '/comments-hack')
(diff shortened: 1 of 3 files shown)


matthew-andrews:
i really didn't want to do this……

client/components/comments/main.js
```diff
@@ -5,7 +5,7 @@ var fetchres = require('fetchres');
 
 module.exports = {};
 module.exports.init = function(uuid, flags) {
-	if (!flags.get('articleComments').isSwitchedOn) {
+	if (flags.get('articleComments').isSwitchedOff || !document.querySelector('#comments')) {
```
matthew-andrews: isSwitchedOff is deprecated

ironsidevsquincy:
understandable - kept it as simple as possible, just title, standfirst, byline and body. Think it's useful while v2 is still bedding in

matthew-andrews:
then can we put it behind a flag that will go back to the old behaviour at the end of Q2?

ironsidevsquincy:
sure can

matthew-andrews:
perhaps reuse

matthew-andrews:
articleCapiV1Fallback

client/components/comments/main.js
commuterjoy: Old habits die hard

https://github.com/guardian/frontend/blob/master/common/app/common/management.scala#L11

matthew-andrews: https://github.com/Financial-Times/next-feature-flags-client/issues/26

PR #401, opened 2015-04-20
fetchUsefuls

matthew-andrews:


server/app.js
```diff
@@ -19,8 +19,12 @@ app.get('^/:id(' + articleUuidRegex + ')/people', require('./controllers/related
 app.get('^/:id(' + articleUuidRegex + ')/organisations', require('./controllers/related/organisations'));
 app.get('^/:id(' + articleUuidRegex + ')/comments-hack', require('./controllers/comments-hack'));
 
+
+// temporary (ma)
+app.get('/more-on/useful', require('./controllers/useful'));
```
ironsidevsquincy: was wondering if these routes should live under the article itself, like people/organisations/comments-hack, as they are related to the article itself, and doesn't pollute the top level, e.g.

`/7f473d26-be68-11e4-8036-00144feab7de/useful` (POST and GET)
`/7f473d26-be68-11e4-8036-00144feab7de/related`
`/7f473d26-be68-11e4-8036-00144feab7de/more-on?relation=primaryTheme`

ironsidevsquincy:
So this is a kind of crowd-sourced popular?

server/app.js
matthew-andrews: at the moment this is global

matthew-andrews:
it's basically recommended, repacked.

PR #413, opened 2015-04-22
Vertical alignment of table elements

ironsidevsquincy:
### Before

screen shot 2015-04-22 at 11 40 03

### After

screen shot 2015-04-22 at 11 40 08


matthew-andrews:
:+1:

PR #415, opened 2015-04-22
Util class to extract uuid form uri

ironsidevsquincy:


server/utils/extract-uuid.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = function (uri) {
+	return uri.replace(/http:\/\/(?:api|www).ft.com\/[^\/]+\/(.*)/, '$1');
```
matthew-andrews: also sometimes mapi.ft.com

ironsidevsquincy: good point. updated

PR #419, opened 2015-04-23
Sepcific styling for images in tables

ironsidevsquincy:
fixes https://github.com/Financial-Times/next-grumman/issues/418

### DIFF
--- a/client/components/article/_main.scss
+++ b/client/components/article/_main.scss
@@ -349,41 +349,4 @@ $header-offset: 30px;
 	margin-bottom: 15px;
 }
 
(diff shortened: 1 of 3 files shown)


matthew-andrews:
:+1:

matthew-andrews:
is `o-table` an option here http://registry.origami.ft.com/components/o-table@1.7.0

ironsidevsquincy:
hmmm, possibly. the current table styling was only meant to be a stop-gap until we designed them properly, so probably worth moving over to `o-table` once we start work on the new design

PR #422, opened 2015-04-23
Now REALLY only show jobs with a start date and no end

ironsidevsquincy:


server/controllers/related/people.js
```diff
@@ -8,7 +8,7 @@ var extractUuid = require('../../utils/extract-uuid');
 
 function getCurrentRole(person) {
 	var currentMembership = _.find(person.memberships, function (membership) {
-		return _.pluck(membership.changeEvents, 'startedAt').shift() && !_.pluck(membership.changeEvents, 'endedAt').shift();
+		return _.find(membership.changeEvents, 'startedAt') && !_.find(membership.changeEvents, 'endedAt');
```
matthew-andrews: if you're using find maybe we can use es6's [`Array.prototype.find`](https://www.npmjs.com/package/array.prototype.find) instead of lodasherscore?  #futureproof

ironsidevsquincy: lodash does have a handy option to pass the property name to search for, but yeah, not a strong reason to use _

matthew-andrews: ah yeah fair enough

PR #424, opened 2015-04-25
Switch to centralised next-build-tools static files build process

matthew-andrews:
cc @commuterjoy, @wheresrhys, @adgad @adambraimbridge @adamsilver @ironsidevsquincy @keirog @aintgoin2goa 

this should cut down the amount of near-identical gulpfile logic in every build and reduces the size & complexity of our Makefiles.


matthew-andrews:
@ironsidevsquincy there is a slight regression on this in that it will no longer notify……  i can add it into the central tools but i heard there was some controversy here……

PR #426, opened 2015-04-27
Only promote links to slideshows to embedded slideshows if they don't ha...

matthew-andrews:
...ve anything inside the <a> tags


matthew-andrews:
cc @ironsidevsquincy

ironsidevsquincy:
:+1:

PR #427, opened 2015-04-27
Valid YouTube video ID

ironsidevsquincy:
Can get `-` and `_`; allow everything to be safe (no reason to lock this down really)


matthew-andrews:
:+1:

PR #428, opened 2015-04-27
Single endpoint for the more ons

ironsidevsquincy:
Means we can de-deupe between more-ons (but doesn't include de-duping story package)

Plus refactor of url formats


server/app.js
```diff
@@ -13,15 +13,14 @@ app.get('/', function(req, res) {
 app.use('^/:id(' + articleUuidRegex + ')', access);
 
 app.get(/^\/fastft\/([0-9]+)(\/[\w\-])?/, require('./controllers/fastft'));
-app.get(/^\/([a-f0-9]+\-[a-f0-9]+\-[a-f0-9]+\-[a-f0-9]+\-[a-f0-9]+)$/, require('./controllers/capi'));
 
+app.get('^/:id(' + articleUuidRegex + ')', require('./controllers/capi'));
```
matthew-andrews: this is much nicer!

matthew-andrews:
:+1:

PR #430, opened 2015-04-28
Barrier fix

aintgoin2goa:
Barrier was appearing underneath the ad instead of next to it, in place of content


matthew-andrews:
ok

PR #432, opened 2015-04-29
If topic is falsey then don't add it to topics array

matthew-andrews:
this is just a bit of a guess here but seems to fix things

@ironsidevsquincy


matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-grumman/group/64253946/

ironsidevsquincy:
yeah, think that's alright :+1: 

Looking at it with fresh eyes, that file's horrible, what with us handling using/not using annotations and now dedpuing (making multiple requests)

matthew-andrews:
:disappointed: 

[insert sweep under the rug gif here]

PR #437, opened 2015-04-30
Correct spacing for more ons

ironsidevsquincy:


client/components/article/_main.scss
```diff
@@ -347,9 +346,8 @@ $header-offset: 30px;
 	line-height: 24px;
 	font-weight: normal;
 }
-
-.article__more-on {
-	margin-bottom: 15px;
+.article__comments > * {
+	margin-top: 15px !important;
```
matthew-andrews: :scream:

ironsidevsquincy: ahh, yeah. due to styling coming from Livefyre, and the css lazy loading. Plus didn't want to use their class names to hook in

PR #441, opened 2015-05-01
protecting against undefined innerText bug in chrome

wheresrhys:
@matthew-andrews

Also scoping all querySelectors


matthew-andrews:
lovely

PR #444, opened 2015-05-07
Handle a person without any memberships

ironsidevsquincy:
Fixes https://app.getsentry.com/nextftcom/ft-next-grumman/group/64722945/


matthew-andrews:
:+1:

PR #445, opened 2015-05-07
More lenient parsing of video id

ironsidevsquincy:
Also, get a rendition which is a closer width

fixes https://app.getsentry.com/nextftcom/ft-next-frontend/group/64572425/


matthew-andrews:
:+1:

PR #447, opened 2015-05-07
Print version

ironsidevsquincy:
Basically remove bumpf

screen shot 2015-05-07 at 14 38 54


client/_print.scss
```diff
@@ -0,0 +1,36 @@
+@media print {
```
matthew-andrews: i'm convince-able either way but feels a little nasty to have so many styles hidden away in a separate file…  just worried people will forget to update this file if they change / remove somethign

ironsidevsquincy: No, good point, and kinda goes against my personal preference of keeping all the styles related to an element in the same place. I'll update

PR #452, opened 2015-05-08
Last defence against crazy-big images

ironsidevsquincy:


matthew-andrews:
i think we should just serve plain text when we fall back to capi v1

matthew-andrews:
(instead of this)

ironsidevsquincy:
means pulling out non-text from the body, which is a bigger piece of work

matthew-andrews:
hmm… there's a quick way to do it

matthew-andrews:
much prefer to be in control of the images we send down to clients…… #453

PR #453, opened 2015-05-08
Use plain text for CAPI v1 fallbacks

matthew-andrews:


matthew-andrews:
Ah i need to switch back to master for next-ft-api-client

matthew-andrews:
@ironsidevsquincy

ironsidevsquincy:
:+1: yep, easier

PR #461, opened 2015-05-14
Make add button work in core experience

adgad:
/cc @ironsidevsquincy @matthew-andrews @wheresrhys 

Note: this means changing the text from "Add to My FT" to "Add" temporarily with the aim of getting all the buttons consistent, so future changes will take place everywhere.


matthew-andrews:
:+1:  sorry for all the failures

PR #470, opened 2015-05-18
Bump to next-sass-setup v6

keirog:
See [sass-setup release notes](https://github.com/Financial-Times/next-sass-setup/releases/tag/v6.0.0).


keirog:
The push build has grumman disease

PR #472, opened 2015-05-18
Anon barrier integration

adamsilver:


aintgoin2goa:
I think @root is probably optional here.

:+1:  (once conflicts resolved and build passes)

PR #479, opened 2015-05-19
Support photo diary images better — treat them as main article images

matthew-andrews:
cc @ironsidevsquincy

### DIFF
--- a/server/controllers/capi.js
+++ b/server/controllers/capi.js
@@ -46,7 +46,7 @@ module.exports = function(req, res, next) {
 
 			var articleV1 = articles[0];
 			var article = articles[1];
-			var $ = bodyTransform(article.bodyXML);
+			var $ = bodyTransform(article.bodyXML, { fullWidthMainImages: res.locals.flags.fullWidthMainImages });
 			var $subheaders = $('.ft-subhead')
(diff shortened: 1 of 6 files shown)


matthew-andrews:
screen shot 2015-05-19 at 17 14 24

server/controllers/capi.js
```diff
@@ -90,7 +90,8 @@ module.exports = function(req, res, next) {
 						isColumnist: isColumnist,
 						// if there's a main image, or slideshow or video, we overlap them on the header
 						headerOverlap:
-							$.root().children('.article__main-image, ft-slideshow:first-child, .article__video-wrapper:first-child').length,
+							$.root().children('.article__main-image, ft-slideshow:first-child, .article__video-wrapper:first-child').length
+							|| $.root().first().children('.article__main-image'),
```
ironsidevsquincy: combine with above? `*:first-child > .article__main-image`

ironsidevsquincy:
:+1:

server/controllers/capi.js
matthew-andrews: hmm it didn't work with `a > .article__main-image` presumably because using `children` rather than `find`

PR #485, opened 2015-05-20
Moved view model to variable

adamsilver:


matthew-andrews:
oh neat, i think i was scared by the phrase “view mode̵l” but this is tidy :+1:

adamsilver:
@matthew-andrews thanks, yep, the suggestion in my gist is for a small abstraction, originally proposed as a constructor (there is method behind madness) but could most definitely just be a function declaration createViewModel() or whatever.  For now this is just fine.

PR #490, opened 2015-05-20
Back to non-ssl

ironsidevsquincy:
IE9 can't handle x-domain requests to a different protocol, e.g. http->https

As we are still serving the site non-securely, doesn't add any extra risk to point these back to http

@richard-still-ft

### DIFF
--- a/client/components/more-on/market-data.js
+++ b/client/components/more-on/market-data.js
@@ -10,7 +10,7 @@ module.exports = function () {
 	var stockPromises = $('.js-markets-data').map(function (el) {
 		var tickerSymbol = el.getAttribute('data-ticker-symbol');
 
-		return fetch('https://ft-next-markets-proxy.global.ssl.fastly.net/securities/v1/quotes?symbols=' + tickerSymbol)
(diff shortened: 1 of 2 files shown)


commuterjoy:
Maybe we should have this going via the router? Or at least via a host name under ft.com

client/components/more-on/market-data.js
```diff
@@ -10,7 +10,7 @@ module.exports = function () {
 	var stockPromises = $('.js-markets-data').map(function (el) {
 		var tickerSymbol = el.getAttribute('data-ticker-symbol');
 
-		return fetch('https://ft-next-markets-proxy.global.ssl.fastly.net/securities/v1/quotes?symbols=' + tickerSymbol)
+		return fetch('http://next-markets-proxy.ft.com.global.prod.fastly.net/securities/v1/quotes?symbols=' + tickerSymbol)
```
matthew-andrews: `//next-markets-proxy`

matthew-andrews:
@captain-sysadmin @commuterjoy can we have `markets-data-proxy-next.ft.com`?

client/components/more-on/market-data.js
ironsidevsquincy: not with that that domain, doesn't handle ssl

PR #502, opened 2015-05-26
More sensible names for article templates, use the save article partial

matthew-andrews:
cc @wheresrhys @adgad

### DIFF
--- a/views/article-v2.html
+++ b/views/article-v2.html
@@ -36,7 +36,7 @@
 						{{/if}}
 						</div>
 						<div class="article__actions" data-o-grid-colspan="12 S6">
-							{{>next-myft-ui/templates/save-for-later contentid=id}}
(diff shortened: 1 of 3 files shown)


views/article-v2.html
```diff
@@ -36,9 +36,7 @@ <h4 class="article__stand-first">{{{articleV1.editorial.standFirst}}}</h4>
 						{{/if}}
 						</div>
 						<div class="article__actions" data-o-grid-colspan="12 S6">
-								{{>next-myft-ui/templates/save-for-later contentId=id}}
-
-							<!-- <a class="article__actions__action" href="">Share</a> -->
+							{{>next-myft-ui/templates/save-for-later contentid=id}}
```
adgad: Is handlebars case sensitive? If so, this should be contentId rather than contentid.

matthew-andrews: oops accident

PR #506, opened 2015-05-27
prevent brands from being removed when more-ons arrive

wheresrhys:
So we have this 
image

instead of this

image

@matthew-andrews @ironsidevsquincy


wheresrhys:
also @toddwebbft @simonwilliamsFT A consequence of the implementation is adding a class to identify that the 'head to head' is a brand. It may be worth adding some design to differentiate from primary theme & section and clear up a little why there are no stories beneath the heading

PR #511, opened 2015-05-28
Rename font mixins

keirog:
…to match [changes in next-type](https://github.com/Financial-Times/next-type/pull/19)


ironsidevsquincy:
:+1:

PR #517, opened 2015-06-02
Stop the 404'ing of comment static assets

matthew-andrews:
Quite a lot of these:-
https://splunk.internal.ft.com/en-GB/app/search/search?q=search%20app%3D%22ft-next-router-v003%22%20status%3D%22404%22%7C%20top%20limit%3D20%20path&earliest=-1d&latest=now&display.general.type=visualizations&display.page.search.tab=visualizations&sid=1433243015.182162.splunk-prod-job-01

### DIFF
--- a/client/main.scss
+++ b/client/main.scss
@@ -11,8 +11,8 @@ $o-gallery-is-silent: false;
 // Fix o-comments at 2.0.3 so we can use the build service for images
 @import "o-assets/main";
 @include oAssetsSetModulePaths((
-	o-comments: '//build.origami.ft.com/files/o-comments@^2.0.0',
-	o-comment-ui: '//build.origami.ft.com/files/o-comment-ui@^2.0.0'
+	o-comments: '//build.origami.ft.com/files/o-comments@2.0.3',
+	o-comment-ui: '//build.origami.ft.com/files/o-comment-ui@2.0.2'
 ));
 $o-comments-is-silent: false;
 @import 'o-comments/main';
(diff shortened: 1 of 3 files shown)


matthew-andrews:
also adds icons that were previously missing

matthew-andrews:
before:-
screen shot 2015-06-02 at 12 17 02

after:-
screen shot 2015-06-02 at 12 17 16

matthew-andrews:
i didn't the js file…… just my editor being helpful……

client/main.scss
```diff
@@ -7,6 +7,13 @@ $o-gallery-is-silent: false;
 @import "o-gallery/main";
 @import "o-fonts/main";
 @include oFontsInclude(MetricWeb, thin);
+
+// Fix o-comments at 2.0.3 so we can use the build service for images
```
ironsidevsquincy: as in, set it to 2.0.3 the `bower.json`?

matthew-andrews: hmm i didn't actually do this did i………  it would be safest to do this but then would need to for o-comments-ui to be a specific version too……

ironsidevsquincy: can we set that in the `resolutions`?

matthew-andrews: it'll probably complain that it's unnecessary…… think it's safer to call it a dependency, which as i'm setting the `oAssets` module path isn't entirely inaccurate

ironsidevsquincy:
:+1:

PR #519, opened 2015-06-03
Use unique syncids per slideshow in an article to support multiple slideshows

matthew-andrews:
should make @musthemag happy

http://next.ft.com/7316cacc-07d2-11e5-a58f-00144feabdc0

{right now when you go forward on one slideshow all others on the same page progress}


matthew-andrews:
cc @ironsidevsquincy

PR #520, opened 2015-06-03
Foot follows

wheresrhys:
@ironsidevsquincy


wheresrhys:
image

client/components/more-on/_main.scss
```diff
@@ -1,105 +1,112 @@
...
-	padding: 5px;
-	font-size: 17px;
-	font-weight: 200;
-	line-height: 1em;
+	margin: 0;
+	padding: 5px 0;
+	@include nTypeAlpha(4);
 	box-sizing: border-box;
 
+	a {
```
ironsidevsquincy: no class name available?

matthew-andrews:
looks aweseom :+1:

client/components/more-on/_main.scss
```diff
@@ -1,105 +1,112 @@
...
+	.more-on--people .more-on__title:before {
+		@include nextIcon(profile, oColorsGetPaletteColor(cold-3), 30);
+	}
+
+	.more-on--regions .more-on__title:before {
+		@include nextIcon(section-world, oColorsGetPaletteColor(cold-3), 30);
+	}
+
+	.more-on--topics .more-on__title:before {
+		@include nextIcon(section-special-report, oColorsGetPaletteColor(cold-3), 30);
```
ironsidevsquincy: seeing all these, I wonder if we should have the ability to use `nextIcon` which just adds the unique styles, as we're repeating things like `display: inline-block` etc for each one of these

wheresrhys: css-clean was sorting this for us, but not sure any more. yes good idea

ironsidevsquincy:
:dancers: :dancers:  :dancers: :dancers:

PR #535, opened 2015-06-12
make sure metadata also returns a content_classification

aintgoin2goa:


matthew-andrews:
how come?

matthew-andrews:
i only recently changed this:-
https://github.com/Financial-Times/next-article/commit/ae42e37ac58a04e110673e9373b5b4a7ca71e91d

matthew-andrews:
if the idea here is to remove those elements from pages that show barriers i think it would be better to stop them from being requested in the first place than make them error

aintgoin2goa:
it's not about barriers - the Access endpoint expects anything with a uuid to respond to a HEAD request to give a content_classification.  I'm seeing requests to access for these specific endpoints take > 8 seconds.  Presumably cos Access is expecting headers and retries / gets confused.

The other option is to adjust the access_metadata file, but that might end up accidentally making content free

matthew-andrews:
i have another idae

aintgoin2goa:
are you going to share it?

matthew-andrews:
https://github.com/Financial-Times/next-article/pull/536

also it might help to move articles to `/content/uuid` (something i planned to do at some point)

PR #536, opened 2015-06-12
Prefix ajax with app name (article) to avoid barrier requests

matthew-andrews:
@aintgoin2goa would this also do the trick?

### DIFF
--- a/client/components/comments/main.js
+++ b/client/components/comments/main.js
@@ -9,7 +9,7 @@ module.exports.init = function(uuid, flags) {
 	if (!flags.get('articleComments') || !document.querySelector('#comments')) {
 		return;
 	}
-	return fetch('/' + uuid + '/comments-hack', { credentials: 'same-origin' })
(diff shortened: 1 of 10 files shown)


aintgoin2goa:
yes - that could work.  wouldn't it be better to use something more semantic though?  Using the app name gives you nothing out-of-the box if that is what you were hoping for.

matthew-andrews:
You get the route sort of for free when you use the app name (ie. already set up in the service registry) so that's the out of the box thing……  Think we should merge this for now and have a think……  Quite like the app name as a namespace for stuff but not that wedded to it.  Even inside this single app we've been a bit inconsistent about this.

It might make more sense if it were /article-page (which it will be eventually)…

aintgoin2goa:
ok - I'll update the access_metadata file.  
Just so we're clear anything prefixed with 'article' will be free

matthew-andrews:
yup

PR #542, opened 2015-06-15
Kill moreOnImages

matthew-andrews:


server/controllers/related/more-on.js
```diff
@@ -67,60 +66,8 @@ module.exports = function (req, res, next) {
...
-				})
-				.filter(_.identity);
-
-			if (!imagePromises.length) {
-				throw new Error('No related');
-			}
-
-			return Promise.all(imagePromises);
-		})
-		.then(function (results) {
```
ironsidevsquincy: jesus, forgot it was this much code

matthew-andrews: a lot, but super stable

PR #543, opened 2015-06-15
Adding scroll depth js

ironsidevsquincy:
Tracks how far users scroll down on aricles


matthew-andrews:
Is it not possible to achieve the same tests with saucelabs?

[We probably need a better way of running them locally……]

matthew-andrews:
Also also @commuterjoy is scroll depth measurements something we want only for articles?

matthew-andrews:
[my second comment — or my first comment — shouldn't stop this being merged……  but we might want to tidy them up in the future]

ironsidevsquincy:
@matthew-andrews, not sure I understand the sauce labs comment...(wrong PR)?

But yeah, now you say it, might make sense to measure this on streams/home page too. Will leave it until someone requests it

matthew-andrews:
sauce labs yes that was mysterious — I mean… instead of adding karma could we add this test to the nightwatch suite that runs in sauce labs?

ironsidevsquincy:
using karma as it's really just a unit test. 

could also add an integration test to sauce labs, i.e. test scroll depth working on an actual next article, but we'd somehow have to be able to monitor events being sent to the beacon endpoint, which I don't think we can do with SL

matthew-andrews:
> test scroll depth working on an actual next article

^^ i'm not sure that's particularly different from what nightwatch is doing

ironsidevsquincy:
yep it is, and see further comment

matthew-andrews:
ah hh ok yes

adambraimbridge:
@matthew-andrews @ironsidevsquincy @commuterjoy It's great to see this in articles, and yes — it'd be good to measure scrolldepth with every beacon event. When we make a scrolldepth tracking more universal, we'll have to take care not to create duplicate/conflicting scrolldepth data.

PR #544, opened 2015-06-16
Trial barrier prep

adamsilver:
* Scaffolding to accomodate new trial barrier
* Turn UI components off when showing a barrier


aintgoin2goa:
:+1:

PR #548, opened 2015-06-16
Use links mixin

keirog:
Link mixins live in [next-sass-setup](https://github.com/Financial-Times/next-sass-setup/blob/master/_links.scss).

Before
before

After
after


ironsidevsquincy:
ohhh, yes

PR #551, opened 2015-06-17
Redirect to interactive based on spreadsheet

ironsidevsquincy:
https://docs.google.com/spreadsheets/d/1Rlx_BfkwCpE8KBJZVK_p5gO61HZF7fRckQige6Hwbl8/edit#gid=0

Also gives the option of embedding the iframe directly in the 'next' page

### DIFF
--- a/server/controllers/interactive.js
+++ b/server/controllers/interactive.js
@@ -1,0 +1,23 @@
+'use strict';
+
+module.exports = function (req, res, next) {
+	// assumes polling has already started
+	var wasRedirected = require('../lib/ig-poller').getData().some(function (mapping) {
+		if (mapping.articleuuid === req.params.id) {
+			if (mapping.displaytype === 'embed') {
+				res.render('interactive', {
+					layout: 'wrapper',
+					interactiveUrl: mapping.interactiveurl
+				});
+			} else {
+				res.redirect(mapping.interactiveurl);
+			}
+			return true;
+		}
+		return false;
+	});
+
+	if (!wasRedirected) {
+		next();
+	}
+};
--- a/server/controllers/interactive.js
+++ b/server/controllers/interactive.js
@@ -4,14 +4,13 @@ module.exports = function (req, res, next) {
 	// assumes polling has already started
 	var wasRedirected = require('../lib/ig-poller').getData().some(function (mapping) {
 		if (mapping.articleuuid === req.params.id) {
-			var interactiveFullUrl = 'http://ig.ft.com' + mapping.interactiveurl;
 			if (mapping.displaytype === 'embed') {
 				res.render('interactive', {
 					layout: 'wrapper',
-					interactiveUrl: interactiveFullUrl
+					interactiveUrl: mapping.interactiveurl
 				});
 			} else {
-				res.redirect(interactiveFullUrl);
+				res.redirect(mapping.interactiveurl);
 			}
 			return true;
 		}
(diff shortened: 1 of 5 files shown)


server/controllers/interactive.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+module.exports = function (req, res, next) {
+	// assumes polling has already started
+	var wasRedirected = require('../lib/ig-poller').getData().some(function (mapping) {
+		if (mapping.articleuuid === req.params.id) {
+			var interactiveFullUrl = 'http://ig.ft.com' + mapping.interactiveurl;
```
matthew-andrews: should we limit only to `ig.ft.com` domains?

matthew-andrews:
:+1:

matthew-andrews:
@kavanagh should have a play with this :-)

server/controllers/interactive.js
ironsidevsquincy: if we don't, people would be opted out, no?

matthew-andrews: i was just thinking out loud whether we should allow any website to be 302'd/proxy'd. but even the ft.com IG links wouldn't opt people out  they would just get bounced about, e.g:-

http://next.ft.com/__opt-in
http://www.ft.com/ig/sites/2015/special-reports/shanghai/

matthew-andrews:
i've tried to improve the nightwatch tests so hopefully wont see this failure again

kavanagh:
Wait, let me get up to date with this.... this is what I think you've done:

1. when we create link files in Methode we also enter the UUID into this spreadsheet (because link files are not a thing in CAPI)
2. when you get a request for a UUID that's not in CAPI you then lookup to see if that's a UUID in the spreadsheet
3. You then either redirect to the URL the browser of embed it with an iframe filling the page.

Is that correct?

Sounds good.

What's the rationale behind the iframe bit?

kavanagh:
ps I don't have access to the spreadsheet so trying to figure it our based on the Bertha response.

matthew-andrews:
Yes the workflow above would work but the intention is for it to be more versatile than that…

- You could, if you wanted, author the content that would be in the interactive graphic in Methode *properly* so that it *does* come through the API.  Then the IG itself could be responsible for loading & displaying that content.  This would also mean all the content in microsites built in this way would be available in search and have proper metadata associated with it such that the underlying articles would appear on stream pages but they would never render as standard articles on Next.
- Then for each article of the type above you would enter the UUIDs into the spreadsheet (which you now have access to).  If the IG supported `#` anchors you could even have one article per chapter, say, and link directly to it.
- The rational for the iframe is simply that we want to provide as much flexibility as possible……  Once this PR is merged you would have two options — either making those articles 302 to the IG or proxying through to Next via an iframe.  In the future we could even proxy directly onto Next on the server side.  (This would require IG's to have absolute paths to its assets or some magic on our side)

matthew-andrews:
Ultimately the goal is enable Next to help IG solve the problem that the content within Interactive Graphics don't appear on Next's metadata-driven stream pages (to come through, the content *needs* to be real Content API objects — not link files); don't work with Next's follow and don't appear in search and in general don't work like other articles on FT.com.  (The proposal is making work like normal articles simply by making them actual FT articles in every way except for render)

server/controllers/interactive.js
ironsidevsquincy: good point. I'll change so the full url is entered into the spreadsheet, more flexible

PR #559, opened 2015-06-18
Catch errors in more ons

ironsidevsquincy:
This endpoint is seriously slow, times out most of the time

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "ft-next-article-primary-tag": "^1.0.0",
     "ft-next-express": "^11.0.0",
     "ft-next-logger": "^1.1.0",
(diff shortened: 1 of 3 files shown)


matthew-andrews:
hmm… don't like suppressing all errors blindly.

should be able to do a simple `instanceof` check here…… next-ft-api-client will convert timeouts to not `.ok` responses

ironsidevsquincy:
k, updated to only ignore `ReadTimeoutError`s, though thinking about it, should probably send these to splunk

ironsidevsquincy:
got some splunk logging now

matthew-andrews:
:+1:

PR #566, opened 2015-06-22
Tidy up transforming

ironsidevsquincy:
For now, just makes the transform chain more sane - https://github.com/Financial-Times/next-article/blob/transform-tidy/server/transforms/body.js#L42-L57

This might become moot once we move over to XSLT, but might be able to keep the same transforming style

cc: @i-like-robots


matthew-andrews:
@ironsidevsquincy @i-like-robots you might like to add some image diff captures for some of the things we're currently transforming so we can visually check the refactors are having no impact :)

matthew-andrews:
https://github.com/Financial-Times/next-article/blob/master/test/visual/test.js#L20-26

ironsidevsquincy:
good idea. I'll add some tests for the components

matthew-andrews:
just to warn you — images typically don't work well…… I might need to add some logic to hide images.  poke me if this causes you pain and i'll make that tweak

ironsidevsquincy:
should be fine without images. why don't they work? because the image diff uses compressed images to compare?

matthew-andrews:
i don't know… i haven't had a chance to debug it yet :smile: … if you'd like to look into it, would be great :-)

PR #568, opened 2015-06-22
Strip out http://www.ft.com.thing/ from IDs

matthew-andrews:
cc @ironsidevsquincy 

@martingartonft  :wink:

### DIFF
--- a/server/controllers/related/more-on.js
+++ b/server/controllers/related/more-on.js
@@ -108,11 +108,16 @@ module.exports = function (req, res, next) {
 						.slice(0, index)
 						.map(function(result) { return result[0]; }));
 					// dedupe
-					var dedupedArticles = _.filter(articleModels, function(articleModel) {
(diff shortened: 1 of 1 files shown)


server/controllers/related/more-on.js
```diff
@@ -108,11 +108,16 @@ module.exports = function (req, res, next) {
...
-							return otherArticleModel.id === articleModel.id;
+					var dedupedArticles = articleModels
+						.map(function(articleModel) {
+							articleModel.id = articleModel.id.replace('http://www.ft.com/thing/', '');
+							return articleModel;
+						})
+						.filter(function(articleModel) {
+							return !otherArticleModels.find(function(otherArticleModel) {
+								return otherArticleModel.id === articleModel.id;
+							});
```
ironsidevsquincy: shouldn't the id changing happen last, or else this filter will never match?

PR #575, opened 2015-06-23
visual diffs only work with core js features, and slideshow is not av…

matthew-andrews:
…ailable at all in core for various reasons


ironsidevsquincy:
:+1:, thanks

PR #576, opened 2015-06-23
Bumped the comments version again

richard-still-ft:


bower.json
```diff
@@ -11,7 +11,7 @@
     "o-gallery": "^1.6.1",
     "o-date": "^1.1.1",
     "n-video": "^2.5.1",
-    "o-comments": "2.0.5",
+    "o-comments": "2.0.6",
```
ironsidevsquincy: any reason this is fixed?

richard-still-ft: You mean the version? Not that I know of... I assumed there was so left it that way but it would nice to allow patch versions.

matthew-andrews: yes there is

matthew-andrews: *needs* to be kept in sync with this here:-
https://github.com/Financial-Times/next-article/blob/132588b6f756cd701b4b396fa4e06349d35c4c46/client/main.scss#L18-20

ironsidevsquincy: oh yeah, that's right. icky

matthew-andrews:
!!!

matthew-andrews:
you didn't fix it

matthew-andrews:
https://github.com/Financial-Times/next-article/pull/577

PR #592, opened 2015-06-26
Use Financier for article copy (internal experiment)

keirog:
Behind "serification" flag


matthew-andrews:
ahaha nice

commuterjoy:
+1 #fontgate

keirog:
flag off:
off

flag on:
on

PR #594, opened 2015-06-26
Use cards for more-ons

keirog:


keirog:
before
before

after
after

keirog:
Let's see if the screenshot bot agrees…

keirog:
(Also this is just step 1 of a few more-on/related updates)

matthew-andrews:
i don't think imagediff  bot covers this because morons are an enhanced experience only feature — currently not coverd by imagediff bot.

server/controllers/related/more-on.js
```diff
@@ -107,6 +107,14 @@ module.exports = function (req, res, next) {
 						.slice(0, index)
 						.map(function(result) { return result[0]; }));
 
+					// add props for more-on cards
+					articleModels.forEach(function (articleModel) {
+						articleModel.headline = articleModel.title;
+						articleModel.lastUpdated = articleModel.publishedDate;
+						// Use bare card type
+						articleModel.isBare = true;
```
matthew-andrews: woah magic.  i didn't know you could do this.  is there any benefit to doing this over a `map`?  i guess less lines of code.

matthew-andrews:
:+1:

server/controllers/related/more-on.js
keirog: for me, forEach seems to make the intention clearer: we're decorating the model collection rather than creating a new one

PR #597, opened 2015-06-26
Do not attempt to transform or de-xmlify an article when logged out

i-like-robots:


ironsidevsquincy:
:+1:

matthew-andrews:
Don't let this stop merging but we should get a simple mocha unit test to cover this :-)

Issue #510, opened 2015-05-28
Slightly odd paragraph spacing


matthew-andrews:
A nice little issue for bug duty person today perhaps, @phamann ?

![screenshot_2015-05-27-23-44-06](https://cloud.githubusercontent.com/assets/825088/7853959/858ba6fa-050c-11e5-96bd-7f6b2420d5fd.png)

reported by @robertshilston

PR #598, opened 2015-06-29
story package timestamp para spacing fix

andygnewman:
Fixed issue documented here https://github.com/Financial-Times/next-article/issues/510

@matthew-andrews - this is a fix to an issue you raised - played around with squashing previous branch commit history, but was still messy, so started from scratch as it was a 2 line fix. Will close previous pr on this.

Fixed in CSS;
- removing float: left from timestamp to get timestamp within overall container
- reducing timestamp top/bottom margins from 10px to 8px to ensure box is same height as image (56px) when story headline on single line

Before
screen shot 2015-06-26 at 11 45 44

After
screen shot 2015-06-26 at 11 46 07

### DIFF
--- a/client/components/_story-package.scss
+++ b/client/components/_story-package.scss
@@ -37,9 +37,6 @@
 	}
 	&:first-child {
 		margin-top: 0;
-		.story-package__article__timestamp {
-			float: left;
-		}
 		.story-package__article__link {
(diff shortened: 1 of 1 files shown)


ironsidevsquincy:
:+1:

PR #603, opened 2015-06-29
Use FinancierTextWeb for body text, in serification test

keirog:
…but keep Financier Display for headline


keirog:
Difficult to show here, but:
Before
before

After
after

matthew-andrews:
Save | Comments a bit funky……

keirog:
Yep, but nothing to do with this change, no?

matthew-andrews:
nope :+1:

keirog:
Ta. (That should change soon anyway, with new button coming in.)

PR #607, opened 2015-06-30
Start moving the transforms over to XSLT

ironsidevsquincy:
cc @i-like-robots


i-like-robots:
This looks really good! Pass in the flags then :+1:

PR #626, opened 2015-07-01
full width inline images

i-like-robots:
Allows authors to mark images to be displayed full width (by appending `{L}` to the the image meta data alt text).

screen shot 2015-07-02 at 10 15 15

re. https://github.com/Financial-Times/next-feature-flags-api/pull/210


matthew-andrews:
:+1:

PR #628, opened 2015-07-01
Loosen limits on Node version

matthew-andrews:


matthew-andrews:
seems the lintspace bug still persists

PR #630, opened 2015-07-01
Strip primary tag from tags

keirog:


next-team:
I've spotted a [difference](https://s3-eu-west-1.amazonaws.com/ft-next-qa/image_diffs/article/2015-07-01/18:01-2375/failures/index.html) between your pull request and production — [see full output](https://s3-eu-west-1.amazonaws.com/ft-next-qa/image_diffs/article/2015-07-01/18:01-2375/successes/index.html)

matthew-andrews:
what does this do?

keirog:
Ah, it strips out the primary tag if it exists in the tags collection, so that the "mentioned in this article" list doesn't show it again. It's already being shown above the headline. This was requested by Dan and Simon. Seemed fair enough!

It turns this:
screen shot 2015-07-02 at 10 10 37
Into this:
screen shot 2015-07-02 at 10 10 29

It keeps the real primary tag in the list if the displayed "primaryTag" (i.e. output of next-article-primary-tag) happens to be the brand instead.
So this…
screen shot 2015-07-02 at 10 10 59
…stays the same, despite Greece being the real primary tag in the content.

Hope that makes sense.

matthew-andrews:
makes sense :)

matthew-andrews:
:+1:

PR #632, opened 2015-07-02
Don't move images out of <p> tags if that <p> also contains text

matthew-andrews:
- Also add test to ensure this never happens again.

### DIFF
--- a/server/stylesheets/images.xsl
+++ b/server/stylesheets/images.xsl
@@ -2,7 +2,7 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
 
     <!-- just moves images out of their containing p -->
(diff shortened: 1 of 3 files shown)


server/stylesheets/images.xsl
```diff
@@ -2,7 +2,7 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
 
     <!-- just moves images out of their containing p -->
-    <xsl:template match="/body/p[ft-content[@type='http://www.ft.com/ontology/content/ImageSet'] and count(*) = 1]">
+    <xsl:template match="/body/p[ft-content[@type='http://www.ft.com/ontology/content/ImageSet'] and count(*) = 1 and not(text())]">
```
i-like-robots: I think you can do `...[ft-content[contains(@type, 'ImageSet')] and...` which is very slightly shorter.

Might need to also check that `text()` only returns actual text and not whitespace etc (`text()[normalize-space()]`) but I'm not sure.

i-like-robots:
:+1:

PR #635, opened 2015-07-02
Fix styling of myFT component

phamann:
### Before
google chrome

### After
google chrome

/cc @adgad @i-like-robots


i-like-robots:
:ship:

matthew-andrews:
happy days

PR #639, opened 2015-07-02
Add mechanism to test typography variants with customers

keirog:
Provides a way for us to show different typography variants to customers.

The (hacky and inelegant) mechanism works via hash in the URL:
http://next.ft.com/[article-id]#typog-switcher:variant-[0-9]
This'll add a class to the body, to give a hook for varying CSS styles.
The variant is saved in localStorage to persist the styling for the customer.

Note: Only runs if the typographySwitcher flag is switched on, and obviously only works for enhanced-experience users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "ft-poller": "0.0.4",
     "lodash": "^3.6.0",
     "merge": "^1.2.0",
-    "next-ft-api-client": "^3.0.0"
+    "next-ft-api-client": "^3.0.0",
+    "superstore": "^1.0.0"
   },
   "devDependencies": {
     "chai": "^2.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -23,7 +23,8 @@
     "ft-poller": "0.0.4",
     "lodash": "^3.6.0",
     "merge": "^1.2.0",
-    "next-ft-api-client": "^3.0.0"
+    "next-ft-api-client": "^3.0.0",
+    "superstore": "^1.0.0"
```
matthew-andrews: lol whattt

keirog: Oh, bad? Seemed better than naked localStorage

matthew-andrews: haha no it is.  just a blast from the past.

matthew-andrews:
:+1:

PR #651, opened 2015-07-07
Genius comments

aintgoin2goa:


commuterjoy:
Nice work.

- Can you make some demo's of the different usage of this? - conversation, footnotes,  
- And show it to Lisa - see if she has ideas on it's usefulness, ideas where to take it next?
- Would be good to think about mobile ux?

PR #661, opened 2015-07-09
add data attributes to barriers

aintgoin2goa:


adamsilver:
Did you want to differentiate between simple and grid?

aintgoin2goa:
no - that's whole point of this.  So we can see which type, of barrier is being not regardless of varient

PR #670, opened 2015-07-10
Whatsapp sharing

adgad:
/cc @commuterjoy @aintgoin2goa 

* Show whatsapp share button on iphone/android
* Lots of hacking around oShare to make it eventually compatible kinda

screenshot_20150710-145055


client/components/share/main.scss
```diff
@@ -0,0 +1,41 @@
+@import "o-share/main";
+
+$_o-share-icon-size: 32px;
+$_o-share-sprite-elements: (twitter, facebook, linkedin, whatsapp);
+
+@include oShareBase(article-share);
+@include oShareActionIcon(facebook, article-share);
+@include oShareActionIcon(twitter, article-share);
+@include oShareActionIcon(whatsapp, article-share);
```
commuterjoy: hide on large viewports?

adgad: Could do. Either way is quite crap - and the whatsapp share doesn't have any fallback if the app doesn't exist.

But actually hiding on > S is less code so I might just do that.

PR #675, opened 2015-07-13
Remove 'meta' property that doesn't seem to be used by anything

matthew-andrews:
@ironsidevsquincy @i-like-robots


i-like-robots:
:ship:

PR #679, opened 2015-07-13
fix issue where text in same p as slideshow got removed

andygnewman:
Fix issue raised where text in same p tag as a slideshow got removed by the xsl process.
Unit test also written to cover this scenario.

### DIFF
--- a/test/server/stylesheets/slideshow.test.js
+++ b/test/server/stylesheets/slideshow.test.js
@@ -62,4 +62,23 @@ describe('Slideshow', function () {
 			});
 	});
 
+	it('should retain any text in the same <p> tag as the slideshow in a separate <p>', function() {
+		return transform(
+				'<body>' +
+					'<p>' +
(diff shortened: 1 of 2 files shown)


test/server/stylesheets/slideshow.test.js
```diff
@@ -62,4 +62,23 @@ describe('Slideshow', function () {
 			});
 	});
 
+	it('should retain any text in the same <p> tag as the slideshow in a separate <p>', function() {
+		return transform(
+				'<body>' +
+					'<p>' +
+						'<a href="http://www.ft.com/cms/s/0/f3970f88-0475-11df-8603-00144feabdc0.html#slide0"></a>' +
+						'Some text in the same p tag as the slideshow' +
```
matthew-andrews: what happens if they put a link or bold text here?

andygnewman: @matthew-andrews neither Matt H nor I can solve your extended problem, suggest you +1 this solution which solves the identified issue and I stackoverflow a more robust solution.

matthew-andrews: definitely worth merging this as it's an improvement……

Issue #681, opened 2015-07-13
Scroll depth triggered JS error when no article [ barrier displayed ]


aintgoin2goa:
<img width="1273" alt="screen shot 2015-07-13 at 12 25 08" src="https://cloud.githubusercontent.com/assets/18816/8648666/535deee6-295a-11e5-8022-d3ba0aa15df2.png">

mr hurley would you mind having a look at this?

https://app.getsentry.com/nextftcom/ft-next-frontend/group/64114384/

aintgoin2goa:
[posted by matt a]

PR #682, opened 2015-07-13
Only run article scroll depth if there's an article body

ironsidevsquincy:
i.e. not a barrier - fixes https://github.com/Financial-Times/next-article/issues/681

Also fix some sass lint warnings


client/components/article/scroll-depth.js
```diff
@@ -33,7 +33,7 @@ var scrollDepth = {
 		opts = opts || {};
 		// allow mocking of window height
 		mockedWindowHeight = opts.windowHeight;
-		if (flags.get('articleScrollDepthTracking') || true) {
+		if (flags.get('articleScrollDepthTracking') && document.querySelector('.article__body')) {
```
ironsidevsquincy: cc @aintgoin2goa

PR #683, opened 2015-07-14
Slideshow - Bugfix/missing para

andygnewman:
Add edge cases to previous fix, so that any other tagged elements within the <p> element hosting the slideshow will also be retained.


matthew-andrews:
:+1: 

hack hack hack :smile:

PR #696, opened 2015-07-15
fix article body strong overwriting font for quotes

andygnewman:
Fix this issue raised through [next feedback] (https://groups.google.com/a/ft.com/forum/#!topic/next.feedback/RdZcufsNwmQ)


client/components/article/_main.scss
```diff
@@ -179,7 +179,7 @@ $header-offset: 30px;
 		margin-top: 0;
 	}
 
-	strong {
+	strong:not(.n-quote >) {
```
ironsidevsquincy: wont work in IE < 9 (http://caniuse.com/#search=%3Anot) - does it look ok in IE8

PR #701, opened 2015-07-16
undo space encoding before reencoding for image service

andygnewman:
Many images from Clamo (fastFT CMS) have spaces in file names. XSLT is encoding these spaces to %20 and then our resize service is encoding the whole url of the file so the %20 becomes %2520 which fails when processed by the responsive image webservice.

This 'decodes' the XSLT partial encoding of the url, before the whole of it is re-encoded.


commuterjoy:
good spot +1

Issue #650, opened 2015-07-07
Ads and related article weirdly aligns……


matthew-andrews:
![screen shot 2015-07-07 at 14 52 57](https://cloud.githubusercontent.com/assets/825088/8547595/e79bae24-24b7-11e5-82dc-a4eb5f510517.png)

matthew-andrews:
ipad portrait i believ

andygnewman:
I think o-ads could do with a clear left to stop this from happening - will take it up with origami team

PR #705, opened 2015-07-16
ad unit clear left

andygnewman:
@matthew-andrews to resolve this issue https://github.com/Financial-Times/next-article/issues/650


matthew-andrews:
ah cool, neat :+1:

PR #709, opened 2015-07-17
Remove dependency on FT.com for whether to show comments

matthew-andrews:


matthew-andrews:
This also enables enables comments for blog articles :smile:

PR #711, opened 2015-07-20
Bug/fix promobox images

andygnewman:
Transfer of images transform to xslt meant that promo-box image classes were not being appended. This solves that issue for the new promo-box style (currently switched on behind flag). Will need to make happen on old promo-box style as well.

### DIFF
--- a/server/transforms/images.js
+++ b/server/transforms/images.js
@@ -29,7 +29,7 @@ module.exports = function($body, opts) {
 				var $img = $body('img[data-image-set-id="' + imageSet.id.replace('http://www.ft.com/thing/', '') + '"]')
 					.attr('src', imgSrc);
 
-				if (imageSet.title) {
+				if ($img.parent().filter('figure').length > 0 && imageSet.title) {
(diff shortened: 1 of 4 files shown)


matthew-andrews:
the article transform logic is getting rather complicated … i wonder if we should for the format that we get to be simplified.

thoughts?  @andygnewman @ironsidevsquincy @i-like-robots ?

server/transforms/images.js
```diff
@@ -29,7 +29,7 @@ module.exports = function($body, opts) {
 				var $img = $body('img[data-image-set-id="' + imageSet.id.replace('http://www.ft.com/thing/', '') + '"]')
 					.attr('src', imgSrc);
 
-				if (imageSet.title) {
+				if ($img.parent().filter('figure').length > 0 && imageSet.title) {
```
i-like-robots: Does `.parent()` accept a selector as a filter?

andygnewman: @i-like-robots Unfortunately not.  .children() does, but not .parent()

i-like-robots:
:ship:

PR #712, opened 2015-07-20
Fix images within promoboxes and tables

andygnewman:
Add logic into images transform to not add inline / pull out classes to images in article that are contained in something other than a <p>, eg. promo-boxes and tables.

From this
screen shot 2015-07-20 at 15 24 54

to this
screen shot 2015-07-20 at 15 25 26

Plus added tests and changed some logic in promo-boxes transform that had fixed it just for promo-boxes.


i-like-robots:
Nice :ship:

PR #714, opened 2015-07-21
Delete advanced list transforms and styles

i-like-robots:
The flag has expired. These were proof of concept and far too complex to be implemented in a non-experimental capacity.

See also: https://github.com/Financial-Times/next-feature-flags-api/pull/246


andygnewman:
+1

PR #715, opened 2015-07-21
add classification for fastft articles

aintgoin2goa:


matthew-andrews:
not sure this will work…

matthew-andrews:
oh no you're right wow, had no idea

server/utils/access.js
```diff
@@ -39,6 +39,8 @@ module.exports = function(req, res, next) {
 						case '3' :
 							classification = 'conditional_premium';
 					}
+				}else if(article.item.location.uri.indexOf('fastft') > -1){
+					classification = 'conditional_registered';
```
matthew-andrews: wouldn't we want this to be `conditional_standard`?

matthew-andrews:
:+1:

server/utils/access.js
aintgoin2goa: Current ft.com barrier says "Fast FT is only available to FT.com subscribers and registered users"

PR #718, opened 2015-07-22
photo diary initial image as main image

andygnewman:
Ensure that photo on photo diary article page renders as main image rather than pull out.

Currently;
screen shot 2015-07-22 at 08 36 49

With fix;
screen shot 2015-07-22 at 08 36 26


i-like-robots:
:+1:

PR #719, opened 2015-07-22
Fix barrier upgrade link

adgad:
/cc @adamsilver @aintgoin2goa @matthew-andrews 

(Can't test cause the barrier doesn't show for me locally)


adamsilver:
:+1:

PR #720, opened 2015-07-22
Supress comment notifications since it clashes with myft blob

adgad:
/cc @schmodt @matthew-andrews @ironsidevsquincy (...and rest of Team Engagement)


matthew-andrews:
is there not an option to hide this?

adgad:
Nope, spoke to Roland and he says there's no way to turn it off from Livefyre.

PR #721, opened 2015-07-22
Bug/image encoding

andygnewman:
@i-like-robots 

Not pretty, am sure it can be optimised a bit.

Resolves the double encoding of spaces, but still left with the issue that clamo is not sending a Content Type of image in the header response, so image service won't process.

### DIFF
--- a/server/transforms/external-images.js
+++ b/server/transforms/external-images.js
@@ -1,0 +1,14 @@
+'use strict';
+
+var cheerio = require('cheerio');
+
+module.exports = function($) {
+	$('img[src]').replaceWith(function(index, el) {
+		var $el = cheerio(el).clone();
+		var matcher = /^https:\/\/next-geebee.ft.com\/image\/v1\/images\/raw\/(.+)\?/;
+		var externalURI = $el.attr('src').match(matcher);
+		externalURI && $el.attr('src', $el.attr('src').replace(externalURI[1], decodeURIComponent(externalURI[1])));
+		return $el;
+	});
+	return $;
+};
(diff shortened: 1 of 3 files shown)


server/transforms/external-images.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var cheerio = require('cheerio');
+
+module.exports = function($) {
+	$('img').replaceWith(function(index, el) {
```
i-like-robots: `img[src]`

server/transforms/external-images.js
```diff
@@ -0,0 +1,14 @@
...
+		var $el = cheerio(el).clone();
+		var matcher = /(?:^https:\/\/next-geebee.ft.com\/image\/v1\/images\/raw\/)(?:([\S]+))(?:[?])/;
+		var externalURI = $el.attr('src').match(matcher);
+		$el.attr('src', $el.attr('src').replace(externalURI[1], decodeURIComponent(externalURI[1])));
```
i-like-robots: `externalURI && $el...`

server/transforms/external-images.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var cheerio = require('cheerio');
+
+module.exports = function($) {
+	$('img').replaceWith(function(index, el) {
+		var $el = cheerio(el).clone();
+		var matcher = /(?:^https:\/\/next-geebee.ft.com\/image\/v1\/images\/raw\/)(?:([\S]+))(?:[?])/;
```
i-like-robots: `/^https:\/\/next-geebee.ft.com\/image\/v1\/images\/raw\/(.+)\?/`

i-like-robots:
squerge :+1:

PR #724, opened 2015-07-22
Re-factor images XSL into internal and external stylesheets

i-like-robots:
Slightly less DRY but more explicit.


andygnewman:
LGTM

PR #725, opened 2015-07-22
re-encode the decode external uri

andygnewman:
Fix for encoding of market data charts on top of fix for clamo files with spaces.


i-like-robots:
:ship:

PR #726, opened 2015-07-22
Switching from (Travis + Haikro) to (Circle + Docker)

JakeChampion:


Makefile
```diff
@@ -36,7 +36,7 @@ clean:
 deploy:
 	nbt configure
 	nbt deploy-hashed-assets
-	nbt deploy
+	nbt deploy --docker
 	nbt scale
 
 visual:
```
matthew-andrews: going to need a circle replacement for `TRAVIS_PULL_REQUEST` check here :grin:

Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+
+EXPOSE 3000
+
+ONBUILD COPY . /app/src
+
+ONBUILD USER root
+ONBUILD RUN chown -R app /app
+ONBUILD USER app
+
+ONBUILD RUN npm prune --production; npm prune --production
```
JakeChampion: There must be a better way to resolve the issue of unlinking karma

PR #727, opened 2015-07-23
double encoding of spaces in external url

andygnewman:
@richard-still-ft 
This should fix issue with getting both clamo and FTMarkets external images back from the image service by double encoding spaces in file names.


richard-still-ft:
+1

PR #729, opened 2015-07-23
Make share text smaller

keirog:
Makes things look less lost when the new more-ons come in.


keirog:
Before:
share-before

After:
share-after

PR #731, opened 2015-07-23
New more-on design

keirog:


keirog:
Before (mobile) | After (mobile)
--- | ---
more-on-mobile-before | more-on-mobile-after

Before (desktop):
more-on-desktop-before

After (desktop):
more-on-desktop-after

PR #732, opened 2015-07-24
promobox migrated to xslt

andygnewman:
Migrates from cheerio to xslt.

Cheerio code removed.
Flag for new vs old style maintained and all tests migrated.
Removed dependence on images transform adding class names for promoboxes.


i-like-robots:
Nested conditional XSLT! Actually quite easy to follow, just a lot of code. Based on the thoroughness of those tests +1!

PR #737, opened 2015-07-27
default background for all article types pb titles

andygnewman:
Moved styling for .article--default .promo-box__title into .promo-box__title to ensure there is a default for all article types (no differentiated styling had been applied for .article--analysis for example). Future-proofs default for any new article types.


i-like-robots:
:ship:

PR #739, opened 2015-07-27
Live Fyre

commuterjoy:
[Something on Live Fyre](http://webpagetest.internal.ft.com/result/150727_R0_2c0073a0aaea5bbce71f1b3f0e144c26/1/details/cached/) in Firefox is hanging the connection until WPT times out after 120s.

The 'load' event doesn't fire. Putting Live Fyre *after* the load event mitigates the immediate problem.

/cc @richard-still-ft

### DIFF
--- a/client/main.js
+++ b/client/main.js
@@ -65,7 +65,6 @@ setup.bootstrap(function(result) {
 	});
 
 	toc.init(flags);
-	comments.init(uuid, flags);
 	oDate.init(document.querySelector('.article'));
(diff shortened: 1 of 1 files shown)


richard-still-ft:
+1, apart from the tests of course...

commuterjoy:
> apart from the tests of course...

damn that inline git editor

richard-still-ft:
:+1:

PR #746, opened 2015-07-30
label -> prefLabel

matthew-andrews:


i-like-robots:
preferable label says who?

:rocket:

matthew-andrews:
rodderz says so rodderz gets

PR #747, opened 2015-07-30
Adding special report stream card to articles

ironsidevsquincy:
behind `specialReportsChallenge` flag

screen shot 2015-07-30 at 16 15 27


matthew-andrews:
so cool!!

server/controllers/related/special-report.js
```diff
@@ -0,0 +1,59 @@
...
+		useElasticSearch: res.locals.flags.elasticSearchItemGet
+	})
+		.then(function (article) {
+			if (article.item.metadata.primarySection.term.taxonomy !== 'specialReports') {
+				return res.status(200).end();
+			}
+			// get the articles in this special report
+			var specialReportId = article.item.metadata.primarySection.term.id;
+			return api.searchLegacy({
+					query: {
```
matthew-andrews: this should be possible to make work without elastic search

ironsidevsquincy: Ahhh, how?

matthew-andrews: think the way i made this was if you used a simple SAPI compatible query, next-ft-api-client would convert that elastic search……
https://github.com/Financial-Times/next-stream-page/blob/master/server/controllers/v1-taxonomy.js#L36
https://github.com/Financial-Times/next-stream-page/blob/master/server/controllers/v1-taxonomy.js#L56-L62

```
query: "specialReportsId: XXX",
```

maybe?

and maybe without `fields: ['_source'],`

ironsidevsquincy: Works, but it only returns the ids rather than the article's data, which would mean more look ups

Also, shouldn't we be moving away from v1 related endpoints?

PR #753, opened 2015-08-03
Ads update

robin-marr-ft:
Reserve space for banner on larger screen sizes.

@matthew-andrews @ironsidevsquincy


matthew-andrews:
:dancers:

PR #759, opened 2015-08-05
genre styling on more-ons

andygnewman:
Pulling through genre styling into the more-ons at the bottom of the article page.

Also resolves duping of articles that was occurring in more-ons.

screen shot 2015-08-06 at 09 25 15

(NB background colour on the screenshot is not what it is on live - local env issue)


matthew-andrews:
these are looking really really smart now.  i really like them.

PR #760, opened 2015-08-06
No to triple braces

keirog:


matthew-andrews:
:+1:

matthew-andrews:
surprised this didn't trigger image diff

ironsidevsquincy:
image diff currently broken as it can't connect to `ft-next-article.herokuapp.com`

matthew-andrews:
ah

PR #768, opened 2015-08-07
fix for main image min height

andygnewman:
To mitigate main image not being returned so text doesn't overlap the header.


ironsidevsquincy:
:+1:

matthew-andrews:
:+1:

PR #775, opened 2015-08-10
Cope with articles in story packaging 404'ing

matthew-andrews:


server/controllers/related/story-package.js
```diff
@@ -35,6 +35,9 @@ module.exports = function(req, res, next) {
 			if (req.query.count) {
 				articles.splice(req.query.count);
 			}
+			articles = articles.filter(function(article) {
+					return article;
+				});
```
matthew-andrews: necessary because of https://github.com/Financial-Times/next-article/pull/775/files#diff-3fb9465fa558d670797eb2ae62829657R26

matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-article/group/75214316/

andygnewman:
Good spot, thanks

PR #777, opened 2015-08-11
Add fallback image for IE9

matthew-andrews:
@tom-parker 
image


matthew-andrews:
but it seems to sort of fix this:
screen shot 2015-08-11 at 09 23 07

PR #786, opened 2015-08-12
testing urls in the real canary app

wheresrhys:
This replicates almost all the existing tests for the controllers, except for the ones which expect a certain content (usually an empty string). What d'you think to removing the bulk of old tests. I'm open to accepting slightly more complex config for `nbt test-urls`

@i-like-robots @andygnewman @ironsidevsquincy


ironsidevsquincy:
Great! Personally think it's fine to keep `test-urls` pretty simple, i.e. this url should return this status code. Having said that, maybe also 'and should return these headers' might be useful too

PR #790, opened 2015-08-13
Correct encoding of image urls

ironsidevsquincy:
Query string was being encoded twice - now all encoding occurs in `transforms/external-images.js`. 

Also, inline images' urls contain html entities - these need to be unencoded before sending to the image service


andygnewman:
nice LGTM

PR #803, opened 2015-08-18
Removed --docker from Makefile

adambraimbridge:


matthew-andrews:
that should do the trick :+1:

adambraimbridge:
Docker is required, it seems, in order to ensure that `libxslt` is installed. So even if removing `--docker` resolves the deployment issue, it'll merely create another.

PR #805, opened 2015-08-18
Fix a thing that will soon break

matthew-andrews:
but right now this will fail the build


adambraimbridge:
Further reading: (for when it breaks)
https://github.com/heroku/heroku-docker/issues/116
https://github.com/Financial-Times/next-build-tools/pull/184/files

PR #806, opened 2015-08-19
Bump express to fix build

phamann:
/cc @adambraimbridge


adambraimbridge:
@phamann I vote we reject this PR as it's not something we need right now.

ironsidevsquincy:
merge or close

matthew-andrews:
shouldn't matter…  just clear the cache rebuild should be eqiuvalent to this change

PR #810, opened 2015-08-20
Store articles read this session in sessionStorage

adgad:
/cc @phamann @wheresrhys @andygnewman @ironsidevsquincy @i-like-robots 

For use by myft-tray (and anyone else who wants to)


adambraimbridge:
+1 Looks good to merge

PR #821, opened 2015-08-24
remove byline when byline is a hyphen

andygnewman:
Fix this;
screen shot 2015-08-24 at 08 29 43

to this

screen shot 2015-08-24 at 08 29 59


matthew-andrews:
hmm this is weird one…  might be worth putting this check `if (byline === '-')` in [harrier](https://github.com/Financial-Times/next-harrier) so that it gets fixed across the site, not just on the article pages?

matthew-andrews:
maybe in here?
https://github.com/Financial-Times/next-harrier/blob/master/server/lib/articleModel.js

andygnewman:
Another example https://next.ft.com/39d2708c-586c-11e5-9846-de406ccb37f2

matthew-andrews:
can we put this logic in harrier?

matthew-andrews:
will be fixed by new es-interface index https://github.com/Financial-Times/next-harrier/pull/164

PR #822, opened 2015-08-24
Try forcing old bower version

adgad:
/cc @ironsidevsquincy Can do this in the meantime, so all the builds aren't blocked whilst we update component dependencies


ironsidevsquincy:
cool :+1:

PR #836, opened 2015-08-26
Poll blogs access metadata

adgad:
/cc @ironsidevsquincy @matthew-andrews


server/lib/blogs-access-poller.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+var pollerData = { access_metadata: [] };
```
ironsidevsquincy: does this mean if a request for blogs is made before the poller has 'warmed up', the request will go through?

adgad: Yes - it relies on preflight giving it a good buffer time...

ironsidevsquincy:
Great, thanks :+1:

PR #845, opened 2015-08-28
Bump express version for n-layout

tom-parker:


i-like-robots:
`--skip-layout-checks` can be re-enabled not NBT has been updated

PR #847, opened 2015-09-01
Remove leftover myFT reading list template

adgad:
/cc @andygnewman @matthew-andrews :speak_no_evil:


matthew-andrews:
:innocent:

PR #852, opened 2015-09-02
Design unification: more ons

ironsidevsquincy:
screen shot 2015-09-02 at 16 07 37


adgad:
Yay for follow buttons!

next-sass-setup has various mixins and variables that may be useful here - particularly `nLinksTopic`, `nLinksHeadline` and some of the use cases down the bottom here https://github.com/Financial-Times/next-sass-setup/blob/master/_colors-use-cases.scss

Open question as to where these should have been documented since nothing is doing a very good job so far - maybe style guide?

andygnewman:
A few questions;

- What happens with Related Story Packages - do these get a similar design treatment?
- In other areas - eg. Hot Topics, where there is a brand (eg. Lex) we've put this in place of the timestamp. Is this something we should do here also?
- Two of the stories don't have a topic, is this a feature of the data or is it in the code?

PR #854, opened 2015-09-03
Fix print CSS

commuterjoy:
- Centralise all the `display: none` stuff
- Hide adverts, misc article components, my-ft that was obscuring the text
- Added notice from ft.com

Tested on Chrome, Safari, Firefox on OSX.

Before

draghi

After

junker


client/main.scss
```diff
@@ -53,3 +53,25 @@ $o-comments-is-silent: false;
 @import "components/typography-switcher/main";
 @import "components/special-report/main";
 @import "components/related-toast/main";
+
+@media print {
+ 
+	body > section,
```
commuterjoy: I feel that one list of stuff you want to hide is simpler.

client/components/article/_main.scss
```diff
@@ -193,13 +196,7 @@
 		}
 	}
 }
-.article__main-image {
```
commuterjoy: Got rid of this. Doubt people are printing the article for the images.

client/components/article/_main.scss
```diff
@@ -84,7 +81,7 @@
 	text-transform: uppercase;
 	@include print {
 		text-indent: -9999px;
-
+		display: block;
```
commuterjoy: text-indent needed this

PR #860, opened 2015-09-07
Quick bug fixes

i-like-robots:
- [x] Only allow external dependencies to increment by patch
- [x] https://app.getsentry.com/nextftcom/ft-next-article/group/79817367/
- [x] https://app.getsentry.com/nextftcom/ft-next-article/group/75681439/


andygnewman:
LGTM

server/stylesheets/image-combos.xsl
```diff
@@ -28,7 +28,7 @@
     </xsl:template>
 
     <xsl:template match="big-number[preceding-sibling::*[1]/ft-content[@type='http://www.ft.com/ontology/content/ImageSet']]" mode="with-image">
-        <xsl:call-template name="big-number" />
+        <xsl:apply-templates select="big-number" />
```
ironsidevsquincy: don't think this will work, it will only match a `big-number` element inside a `big-number` element

i-like-robots: The page in question is [this one](https://next.ft.com/61ab30e6-0fae-11e5-b968-00144feabdc0) which is currently erroring. This is how the page looks with the above, is it OK?:

screen shot 2015-09-07 at 11 43 07

ironsidevsquincy: it's probably failing because there's no template called `big-number`. This change will only match if there's a `big-number` element within a `big-number` element, so will never match

this whole file was for the article combo prototype - I think it's safe to just delete it for now

i-like-robots: Got it - I've removed the unused big number cheerio stuff too.

PR #865, opened 2015-09-08
Add logging for access middlewhere

tom-parker:
@commuterjoy


server/controllers/access.js
```diff
@@ -80,7 +81,7 @@ module.exports = function(req, res, next) {
 				if (access) {
 					classification = access.classification;
 				}
-
+				metrics.count('__debug.accessmiddleware.accessservice.success.count', 1);
```
commuterjoy: you might want to put this on line 58 (at the top of the function).

then add another counter at the bottom, after the `res.status(200)`

PR #866, opened 2015-09-08
Remove typography switcher

i-like-robots:
re. https://github.com/Financial-Times/next-feature-flags-api/pull/313


andygnewman:
:+1:

PR #867, opened 2015-09-08
Fixes https://app.getsentry.com/nextftcom/ft-next-article/group/75015207/

ifyio:
cc/ @ironsidevsquincy 

Fixes https://app.getsentry.com/nextftcom/ft-next-article/group/75015207/

The error was not being properly caught.


server/lib/no-related-results-exception.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+
+module.exports = function NoRelatedResultsException() {
+	this.name = "NoRelatedResultsException";
+};
```
ironsidevsquincy: worth extending `Error`?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#Custom_Error_Types

matthew-andrews: I'm all for extending errors but at the moment it really smells.
https://github.com/matthew-andrews/fetchres/blob/master/src/fetchres.js#L41-L49

And if you are relying on a error being thrown in a dependency it's really bad to use `instanceof` because if the same dependency is pulled in at different versions in your `node_modules` tree (as npm will let you) `instanceof` will _fail_.

Say you have a `module-a` and that relies on `module-b` at `v1.0.0` and module B throws `ModuleBError` and your main app depends on `module-b` at `v2.0.0`

```js
var moduleA = require('module-a');
var moduleB = require('module-b');

try {
  moduleA.doSomethingThatThrowsAModuleBError();
} catch(err) {
  if (err instanceof moduleB.ModuleBError) {
    // You might think that code here will run.  It won't.
  } else {
    // Instead, this will run.
  }
}
```

I would advise against using `instanceof` entirely and use string equality checks as we now recommend in `fetchres`: https://github.com/matthew-andrews/fetchres#usage-node

```js
var moduleA = require('module-a');
var moduleB = require('module-b');

try {
  moduleA.doSomethingThatThrowsAModuleBError();
} catch(err) {
  if (err.name === moduleB.ModuleBError.name) {
    // This code will now correctly run.
  } else {
    // And this code will now correctly not run.
  }
}
```

ifyio:
@ironsidevsquincy @matthew-andrews You both happy with this? I have updated the code based on Matts suggestion.

server/lib/no-related-results-exception.js
```diff
@@ -0,0 +1,9 @@
+"use strict";
+
+function NoRelatedResultsException() {
+	this.name = "NoRelatedResultsException";
+}
+
+NoRelatedResultsException.NAME = "NoRelatedResultsException";
```
ironsidevsquincy: maybe a `const` of the name, that's used in both places and exposed in exports

ifyio: oops. Forgot to reuse `...NAME` in the class itself. Will make that change. I would have used `const` but an error is thrown when it is used in strict mode and linter requires strict mode

ifyio:
@ironsidevsquincy ?

matthew-andrews:
:+1: from me

PR #869, opened 2015-09-08
Reorder middleware

tom-parker:
cc @matthew-andrews @commuterjoy @ironsidevsquincy


ironsidevsquincy:
we're just exposing headlines, images, etc in this ajax-y requests, no article body, so as long as that's ok not to have a barrier around, :+1:

PR #876, opened 2015-09-15
next read - server side version of next read lead article

andygnewman:
lead article read next (behind articleSuggestedRead flag)

**Key features**
*Header*
- remove tags on RHS and replace with lead read next article
- different visual treatment based on whether the article is more recent than main article
- move standfirst from header to after image, before 1st paragraph

*Bottom of article*
- lead read next article inserted after article sharing and before related stories / more-ons 

*Logic for lead read next article*
- first available of the following (in order of compellingness)
    - more recent story on same topic as main article
    - first story in curated story package
    - most recent story on same topic as main article 

Screenshots
screen shot 2015-09-15 at 16 30 15
screen shot 2015-09-15 at 16 30 35
screen shot 2015-09-15 at 16 31 50
screen shot 2015-09-15 at 16 45 38

(lots of fixtures for test stubbing)

### DIFF
--- a/client/components/read-next/main.scss
+++ b/client/components/read-next/main.scss
@@ -1,0 +1,71 @@
+.article__standfirst {
+	@include nTypeBravo(2);
+}
+
+.next-up {
+	position: relative;
+	padding: 10px 10px 10px 25px;
+}
+
+.next-up__header {
+	margin-top: 10px;
+	padding-top: 15px;
+	border: 1px solid getColor('warm-3');
+}
+
+.next-up__intro {
+	@include nTypeFoxtrot(4);
+	position: absolute;
+	top: -20px;
+	left: 15px;
+	margin: 5px 0;
+	padding: 5px 10px;
+	background-color: getColor('warm-5');
+	color: getColor('cold-2');
+	font-weight: 900;
+	.next-up__bottom & {
+		@include nTypeFoxtrotSize(3);
+		left: -5px;
+		top: -25px;
+	}
+}
+
+.next-up__intro__header {
+	background-color: getColor('pink');
+}
+
+.next-up__headline {
+	@include nLinksHeadline;
+	@include nTypeFoxtrot(2);
+	position: relative;
+	color: getColor('cold-1');
+	text-decoration: none;
+	border-bottom: 0;
+	&::before {
+		content: '>';
+		speak: none;
+		position: absolute;
+		top: -2px;
+		left: -15px;
+	}
+}
+
+.next-up__timestamp {
+	@include nTypeAlpha(5);
+	color: oColorsGetColorFor(timestamp, text);
+	text-transform: uppercase;
+	white-space: nowrap;
+}
+
+.next-up__timestamp__more-recent {
+	color: getColor('claret-inverse');
+}
+
+.next-up__topic-link {
+	@include nLinksTopic;
+}
+
+.read-next__container {
+	margin-top: 40px;
+	border-top: 1px solid getColor('warm-3');
+}
--- a/server/controllers/article.js
+++ b/server/controllers/article.js
@@ -17,6 +17,7 @@ var twitterCardSummary = require('../utils/twitter-card').summary;
 var getDfp = require('../utils/get-dfp');
 var getSuggested = require('./article-helpers/suggested');
 var exposeTopic = require('./article-helpers/exposeTopic');
+var readNext = require('../lib/read-next');
 
 module.exports = function(req, res, next) {
 
@@ -84,9 +85,12 @@ module.exports = function(req, res, next) {
 					useBrightcovePlayer: res.locals.flags.brightcovePlayer ? 1 : 0,
 					renderTOC: res.locals.flags.articleTOC ? 1 : 0,
 					fullWidthMainImages: res.locals.flags.fullWidthMainImages ? 1 : 0,
-					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0
+					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0,
+					suggestedRead: res.locals.flags.articleSuggestedRead ? 1 : 0,
+					standFirst: article[0] ? article[0].item.editorial.standFirst : ""
 				}),
-				socialMediaImage(article[1])
+				socialMediaImage(article[1]),
+				res.locals.flags.articleSuggestedRead && article[0] ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : Promise.resolve()
 			]);
 		})
 		.then(function(results) {
@@ -95,6 +99,7 @@ module.exports = function(req, res, next) {
 			var articleV1 = results[0];
 			var article = results[1];
 			var mainImage = results[3];
+			var readNextArticle = results[4];
 
 			var $ = bodyTransform(results[2], res.locals.flags);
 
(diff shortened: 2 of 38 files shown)


client/components/read-next/main.scss
```diff
@@ -0,0 +1,70 @@
...
+	left: 15px;
+	margin: 5px 0;
+	padding: 5px 10px;
+	background-color: getColor('warm-5');
+	color: getColor('cold-2');
+	font-weight: 900;
+	.next-up__bottom & {
+		left: -5px;
+		top: -25px;
+		@include nTypeFoxtrotSize(3);
```
i-like-robots: mixins should be at the top of the ruleset

client/components/read-next/main.scss
```diff
@@ -0,0 +1,70 @@
...
+
+.next-up__headline {
+	@include nLinksHeadline;
+	@include nTypeFoxtrot(2);
+	position: relative;
+	color: getColor('cold-1');
+	text-decoration: none;
+	border-bottom: 0;
+	&::before {
+		content: '>';
```
i-like-robots: speak: none - otherwise this can be read out as 'greater than'

client/components/read-next/main.scss
```diff
@@ -0,0 +1,70 @@
...
+		margin-bottom: -2px;
```
i-like-robots: Without a Y position this could end up anywhere, maybe `bottom: -2px`?

server/controllers/article.js
```diff
@@ -82,9 +83,12 @@ module.exports = function(req, res, next) {
 					useBrightcovePlayer: res.locals.flags.brightcovePlayer ? 1 : 0,
 					renderTOC: res.locals.flags.articleTOC ? 1 : 0,
 					fullWidthMainImages: res.locals.flags.fullWidthMainImages ? 1 : 0,
-					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0
+					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0,
+					suggestedRead: res.locals.flags.articleSuggestedRead ? 1 : 0,
```
i-like-robots: No need to pass flags from the controller, see https://github.com/Financial-Times/next-article/blob/8f02986d844af99c4a063f5bc33e86a3b39b3f44/views/partials/more-ons.html#L7

server/controllers/article.js
```diff
@@ -134,7 +139,7 @@ module.exports = function(req, res, next) {
 						dfp: metadata ? getDfp(metadata.sections) : undefined,
 						visualCat: metadata ? getVisualCategorisation(metadata) : undefined,
 						isSpecialReport: metadata &&
-							metadata.primarySection.term.taxonomy === 'specialReports'
+							metadata.primarySection.term.taxonomy === 'specialReports',
```
i-like-robots: trailing comma

server/controllers/article.js
```diff
@@ -182,6 +187,10 @@ module.exports = function(req, res, next) {
 						viewModel.twitterCard = twitterCardSummary(article, articleV1, mainImage);
 					}
 
+					if(res.locals.flags.articleSuggestedRead) {
```
i-like-robots: space before parens

server/lib/read-next.js
```diff
@@ -0,0 +1,62 @@
...
+module.exports = function(articleV1, useElasticSearch) {
+	var parent = articleV1;
+	var parentLastPublishedDateTime = parent.item.lifecycle.lastPublishDateTime;
+
+	var storyPackagePromise;
+	if (parent.item.package.length) {
+		storyPackagePromise = api.contentLegacy({
+			uuid: parent.item.package[0].id,
+			useElasticSearch: useElasticSearch,
+			type: 'Article'
```
i-like-robots: There is no `type` option

server/lib/read-next.js
```diff
@@ -0,0 +1,62 @@
...
+		});
+	}
+
+	var parentTopic = articleTopicMapping(parent.item.metadata);
+	var topicPromise;
+	if (parentTopic) {
+		topicPromise = api.searchLegacy({
+			query: parentTopic.taxonomy + 'Id:"' + parentTopic.id + '"',
+			count: 2,
+			fields: true,
```
i-like-robots: :heart:

server/lib/read-next.js
```diff
@@ -0,0 +1,62 @@
...
+					return topicArticle.item.id !== parent.item.id;
+				})
+				.slice(0,1);
+			return articlePodMapping(topicArticle[0]);
+		});
+	}
+
+	return Promise.all([
+		Promise.resolve(storyPackagePromise),
+		Promise.resolve(topicPromise)
```
i-like-robots: `storyPackagePromise` and `storyPackagePromise` are already promises, they need to resolve or reject themselves

server/lib/read-next.js
```diff
@@ -0,0 +1,62 @@
...
+		topicArticle.source = "topic";
+		if (new Date(topicArticle.lastUpdated) > new Date(parentLastPublishedDateTime)) {
+			topicArticle.moreRecent = true;
+			return topicArticle;
+		} else if (packageArticle) {
+			packageArticle.source = "package";
+			return packageArticle;
+		} else {
+			return topicArticle;
+		}
```
i-like-robots:
```js
if (packageArticle) {
	packageArticle.source = "package";
	return packageArticle;
}

if (new Date(topicArticle.lastUpdated) > new Date(parentLastPublishedDateTime)) {
	topicArticle.moreRecent = true;
}

return topicArticle;
```

server/mappings/article-topic-mapping.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+module.exports = function(metadata) {
+	if (!metadata) {
+		return;
```
i-like-robots: This function should never be called. If someone does without an arg then it should error.

server/controllers/article.js
andygnewman: This is passing the flag value to the XSLT to determine whether to put the standfirst below the image. In the partial I'm using the presence of the readNextArticle property of the viewModel to drive rendering.

server/lib/read-next.js
andygnewman: Logic / hierarchy that I'm aiming for here is;

Return 1 article based on;

1) if the most recent article on the same topic is more recent than the parent, use this, if not move to 2

2) if there is a story package, use the first story package, if no story package, use the most recent article on same topic

Your suggestion prioritises the story package, even if topic story is more recent, so I think I'm OK to stick with the original?

server/controllers/article.js
```diff
@@ -82,9 +83,12 @@ module.exports = function(req, res, next) {
...
 					renderTOC: res.locals.flags.articleTOC ? 1 : 0,
 					fullWidthMainImages: res.locals.flags.fullWidthMainImages ? 1 : 0,
-					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0
+					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0,
+					suggestedRead: res.locals.flags.articleSuggestedRead ? 1 : 0,
+					standFirst: article[0] ? article[0].item.editorial.standFirst : ""
 				}),
-				socialMediaImage(article[1])
+				socialMediaImage(article[1]),
+				Promise.resolve(res.locals.flags.articleSuggestedRead ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : undefined)
```
i-like-robots:
```js
Promise.resolve(res.locals.flags.articleSuggestedRead ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : undefined)
```

```js
res.locals.flags.articleSuggestedRead ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : Promise.resolve()
```

server/lib/read-next.js
```diff
@@ -0,0 +1,79 @@
+"use strict";
+
+var articlePodMapping = require('../mappings/article-pod-mapping');
+var articleTopicMapping = require('../mappings/article-topic-mapping');
+var api = require('next-ft-api-client');
+
+module.exports = function(articleV1, useElasticSearch) {
+	if (!articleV1) {
+		return;
+	}
```
i-like-robots: remove this condition

server/lib/read-next.js
```diff
@@ -0,0 +1,79 @@
...
+				.slice(0,1);
+			return articlePodMapping(topicArticle[0]);
+		})
+		.catch(function(error) {
+			return ;
+		});
+	}
+
+	return Promise.all([storyPackagePromise,topicPromise])
+	.then(function(results) {
```
i-like-robots: funky formatting ;-)

PR #877, opened 2015-09-16
remove visual test from provision in Makefile

andygnewman:
As visual test suite is adding no current value and is causing build failure issues, propose removing it from the provision step in Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ provision:
 	nbt configure ft-next-article ${TEST_APP} --overrides "NODE_ENV=branch"
 	nbt deploy-hashed-assets
 	nbt deploy ${TEST_APP} --skip-enable-preboot --docker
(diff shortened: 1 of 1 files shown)


andygnewman:
cc: @matthew-andrews @aintgoin2goa 

taken the simple approach of disabling rather than stripping out completely

Makefile
```diff
@@ -53,7 +53,7 @@ provision:
 	nbt configure ft-next-article ${TEST_APP} --overrides "NODE_ENV=branch"
 	nbt deploy-hashed-assets
 	nbt deploy ${TEST_APP} --skip-enable-preboot --docker
-	make -j2 visual smoke
+	make -j2 smoke
```
matthew-andrews: `-j2` no longer necessary

PR #879, opened 2015-09-16
Podcasts MVP

i-like-robots:
screen shot 2015-09-16 at 14 17 19
screen shot 2015-09-16 at 14 18 04


server/controllers/podcast.js
```diff
@@ -0,0 +1,105 @@
...
+
+		return Promise.all([article, related]);
+	}
+
+	function mapRelated(fulfilled) {
+		var article = fulfilled[0];
+		var related = fulfilled[1];
+
+		article.relatedContent = related
+			.map(function(raw) {
```
andygnewman: What if there are no related articles returned?

i-like-robots: next-ft-api-client should always return an array on success but I should capture any errors...

i-like-robots: Updated!

server/utils/external-podcast-links.js
```diff
@@ -0,0 +1,91 @@
...
+		stitcher: feedNames[1] ? 'http://www.stitcher.com/podcast/financial-times/' + feedNames[1] : null,
+		audioboom: feedNames[2] ? 'http://audioboom.com/channel/' + feedNames[2] : null,
+		soundcloud: feedNames[3] ? 'https://soundcloud.com/ft-analysis' + feedNames[3] : null
+	};
+}
+
+module.exports = function externalPodcastLinks(show) {
+	var data;
+
+	switch(show) {
```
imranolas: This looks like static config that could live elsewhere. A object map might be handy for readability and also remove some duplication.

```js
var config = {
  'Copenhagen Summit podcast' : ['', '', '' , ''],
  'FT Alphachat': ['id448302257', 'ft-alphachat', 'ftalphachat', 'ft-alphachat'],
  ...
}

return apply(config[show]);
```

imranolas: In addition, you could have `apply` handle `undefined` allowing you to remove the entries of `['', '', '', '']`.

i-like-robots: I can definitely DRY this out a bit... brb

andygnewman:
:+1:

PR #883, opened 2015-09-16
Stale cache handling

commuterjoy:
Instructs the cache to update the content every 2 minutes but if the origin is down then show stale content for 3 days.

https://docs.fastly.com/guides/performance-tuning/serving-stale-content

/cc @ironsidevsquincy @ironsidevsquincy worth adding to the front-page imho


aintgoin2goa:
:+1: :+1:

PR #884, opened 2015-09-16
`(--docker)--`

matthew-andrews:


matthew-andrews:
@andygnewman @i-like-robots 

this wasn't that hard :-)

i-like-robots:
Looks good, surprised Circle doesn't baulk at not having its own XSLT version.

matthew-andrews:
it's still installing a version of xslt……

matthew-andrews:
i don't know why the `circle.yml` change breaks the build now !

Issue #890, opened 2015-09-21
XSLT transforms in 1st para with slideshow not being applied


andygnewman:
eg. https://next.ft.com/3fbfe3d6-5cf8-11e5-a28b-50226830d644

Links in first para of article with a slideshow are not undergoing the links transforms (presume it would be the same for any other transforms that should apply to that para).

Test for this should be updated to something like;

    it('should retain any <a> tags in the text in the same <p> tag as the slideshow in a separate <p>',   function() {
		return transform(
				'<body>' +
					'<p>' +
						'<a data-asset-type="slideshow" data-embedded="true" href="http://www.ft.com/cms/s/4fd9b5ca-5cea-11e5-9846-de406ccb37f2.html#slide0"></a>' +
						'Abraham Lincoln and Stephen Douglas famously held three-hour debates on the eve of the US civil war. Duration was the only thing the second 2016 ' +
						'<a href="http://blogs.ft.com/the-world/liveblogs/2015-09-15/" title="As it happened: Republican presidential debate - FTcom">Republican debate</a> ' +
						'had in common with those legendary exchanges.' +
					'</p>' +
				'</body>',
				{
					renderSlideshows: 1
				}
			)
			.then(function (transformedXml) {
				transformedXml.should.equal(
					'<body>' +
						'<ft-slideshow data-uuid="f3970f88-0475-11df-8603-00144feabdc0"></ft-slideshow>' +
						'<p>Abraham Lincoln and Stephen Douglas famously held three-hour debates on the eve of the US civil war. Duration was the only thing the second 2016 ' +
						'<a href="http://blogs.ft.com/the-world/liveblogs/2015-09-15/" data-trackable="link" class="article__body__link">Republican debate</a> ' +
						'had in common with those legendary exchanges.' +
					'</p>' +
					'</body>\n'
				);
			});
	});

andygnewman:
Fixed by this https://github.com/Financial-Times/next-article/pull/894

PR #894, opened 2015-09-21
get transforms working on para with slideshow

andygnewman:
Fixes this issue https://github.com/Financial-Times/next-article/issues/890

### DIFF
--- a/server/stylesheets/slideshow.xsl
+++ b/server/stylesheets/slideshow.xsl
@@ -6,7 +6,9 @@
 
         <xsl:if test="normalize-space(string()) != ''">
           <p>
-            <xsl:copy-of select="./node()[not(self::a[substring(@href, string-length(@href) - 6) = '#slide0' and normalize-space(string()) = ''])]" />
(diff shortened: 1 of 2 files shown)


server/stylesheets/slideshow.xsl
```diff
@@ -6,7 +6,9 @@
 
         <xsl:if test="normalize-space(string()) != ''">
           <p>
-            <xsl:copy-of select="./node()[not(self::a[substring(@href, string-length(@href) - 6) = '#slide0' and normalize-space(string()) = ''])]" />
+            <xsl:apply-templates select="./node()[not(self::a[substring(@href, string-length(@href) - 6) = '#slide0' and normalize-space(string()) = ''])]">
+              <xsl:copy-of select="." />
```
ironsidevsquincy: does this do anything? you can only have `sort` and `with-param` child elements 

https://msdn.microsoft.com/en-us/library/ms256184(v=vs.110).aspx

andygnewman: You're right - good spot, the apply templates on it's own solves the issue.

PR #899, opened 2015-09-22
move data-trackable into a element

andygnewman:
Since last week CTA events only fire on <a> and <button> elements.

Share actions had data-trackable on <i> elements and therefore tracking was broken.

This puts the data-trackable into the surrounding <a> element.


ironsidevsquincy:
:+1: thanks!

Issue #900, opened 2015-09-22
Be more discriminate with the image we show in the more on


ironsidevsquincy:
i.e. we just show the first one, https://github.com/Financial-Times/next-article/blob/master/server/controllers/related/more-on.js#L122

which can lead to https://next.ft.com/27f68194-5f71-11e5-a28b-50226830d644

![screen shot 2015-09-22 at 14 52 34](https://cloud.githubusercontent.com/assets/74132/10020210/950da892-6139-11e5-997d-a81c68d582de.jpeg)

PR #901, opened 2015-09-22
Display correct image in more ons

ironsidevsquincy:
fixes https://github.com/Financial-Times/next-article/issues/900


andygnewman:
:+1:  in some other work I'm merging in later I have mapping functions for article and topic to extract the right data for pods/topics/cards etc. So will seek to integrate this into more-ons etc to drive some consistency. I'll sort that out later this week.

ironsidevsquincy:
cool. it is duplicated in at least 3 places

PR #903, opened 2015-09-23
remove related toast code

andygnewman:
Flag for related toast due to expire and user testing unfortunately unsuccessful, so removing code in article.


ironsidevsquincy:
:+1:

PR #905, opened 2015-09-23
Upgrade to o-grid v4

keirog:
Changes:
- Force-bumped next-sass-setup, next-welcome and o-grid
- Amended markup/styling to reflect (breaking) changes in o-grid v4
  - Added `o-grid-container` wrapper around `o-grid-row`s
  - Zeroed-out the left padding of the article aside, as the `oGridRemoveGutters` mixin was removed

Rationale for the bower resolution nudges:
- next-sass-setup & next-welcome had a major bump because of breaking changes in o-grid
- n-layout (where next-sass-setup/next-welcome/o-grid is pulled in for all
page-serving apps) is being pulled in as * for some apps, so bumping
next-sass-setup there to v7 would break some apps
- doing it this way (i.e. forcing versions here) means we can upgrade
one page-serving app at a time, rather than them all breaking at once
- ideally this would not be necessary


keirog:
(Ended up tidying up the n-layout dependencies and bumping that. Phew)

PR #913, opened 2015-09-24
Remove /<uid> route for articles

i-like-robots:


commuterjoy:
:+1: remove the registry route first of course :ambulance:

i-like-robots:
There's a pR for that too :smile:

Issue #892, opened 2015-09-21
Reduce Elasticsearch requests


i-like-robots:
Currently we may hit ES up to 18 times to render a single article. This is caused by 3 factors:- parts of our application are not talking to each other, we munge V1 and V2 content and our compatibility with CAPI for use as a fallback. An example of the requests might look like:

1. Barriers controller requests V1 and V2 article (munging two data sources)
2. Article controller requests V1 and V2 article again (munging and not talking to each other)
3. Multiple 'more on' or related content requests the article again (not talking to each other) then fetches matching article IDs (compatibility with CAPI) then fetches the article data.

The first issue of munging two versions of the same article may be fixed by https://github.com/Financial-Times/next-harrier/issues/122.

Pieces of code not talking to each other may be a symptom of a few problems: Building features too quickly without prior evaluation or perhaps the lack of conventions for code organisation (such as those enforced by a framework).

The compatibility with a fallback to CAPI may be dropped in future as we seek to push better data into a more scalable ES.

PR #918, opened 2015-09-24
Reduce es calls

andygnewman:
Gets rid of at least 5 ES calls in creating Related Stories, More Ons and Special Reports, plus gets rid of a load of spurious processing and re-rendering of headings.

In line with strategy laid out here https://github.com/Financial-Times/next-article/issues/892
cc: @i-like-robots

### DIFF
--- a/client/components/onward-journey/main.js
+++ b/client/components/onward-journey/main.js
@@ -2,7 +2,6 @@
 
 var fetchres = require('fetchres');
 var oDate = require('o-date');
-var myFtUi = require('next-myft-ui');
 
 // Sort of like Promise.all but will be called whether they fail or succeed
 var allSettled = promises => {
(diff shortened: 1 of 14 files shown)


client/components/onward-journey/main.js
```diff
@@ -16,30 +15,18 @@ var allSettled = promises => {
...
 			}
-			if (opts.renderer) {
-				opts.renderer(el, resp);
-			} else {
-				el.innerHTML = resp;
-			}
+			el.innerHTML = resp;
 			oDate.init(el);
 		})
 		.catch(function() {
```
i-like-robots: .catch(

client/components/onward-journey/main.js
```diff
@@ -47,6 +34,22 @@ module.exports.init = flags => {
 	var article = document.querySelector('.article');
 	var articleId = article.getAttribute('data-content-id');
 	var articleSources = article.getAttribute('data-content-sources');
+	var dehydratedMetadata = document.getElementById('dehydrated-metadata');
+	var hydratedMetadata = dehydratedMetadata && JSON.parse(dehydratedMetadata.innerHTML);
+	var storyPackageIds = hydratedMetadata.package && hydratedMetadata.package.length && hydratedMetadata.package.map(function(el) {
```
i-like-robots: No need to check `length`, you may as well map and return an empty array

server/controllers/article.js
```diff
@@ -109,7 +109,15 @@ module.exports = function(req, res, next) {
...
 			}
-
+			//specialReport is a circular - if it exists, delete it before dehydrating it
+			if (metadata && metadata.primarySection && metadata.primarySection.term.specialReport) {
+				delete metadata.primarySection.term.specialReport;
+			}
+			var dehydratedMetadata = {
+				primaryTheme: metadata && metadata.primaryTheme,
+				primarySection: metadata && metadata.primarySection,
+				package: articleV1 && articleV1.item && articleV1.item.package
```
i-like-robots: This could assign the properties with data unusable by the front end, probably better to switch to a ternary statements and assign `null`

PR #930, opened 2015-09-28
Remove dependencies

i-like-robots:


matthew-andrews:
less is more :+1:

i-like-robots:
```
npm ERR! extraneous: es6-promise
```

Even building with no cache :cry:

PR #931, opened 2015-09-28
Next Reads: Design feedback

imranolas:
- [x] Update component copy/wording
- [x] Display 3 items for read next
- [x] Fix styling for header and bottom next read components
- [x] Hide all more ons when displaying next reads

### DIFF
--- a/client/components/read-next/main.scss
+++ b/client/components/read-next/main.scss
@@ -4,30 +4,53 @@
 
 .next-up {
 	position: relative;
-	padding: 10px 10px 10px 25px;
+	padding: 15px;
+	padding-right: 25px;
 }
 
 .next-up__header {
 	margin-top: 10px;
 	padding-top: 15px;
-	border: 1px solid getColor('warm-3');
+	border: 1px solid #ddccb6;
+	border-width: 1px 0 0 1px;
 }
 
 .next-up__intro {
 	@include nTypeFoxtrot(4);
 	position: absolute;
-	top: -20px;
-	left: 15px;
-	margin: 5px 0;
-	padding: 5px 10px;
 	background-color: getColor('warm-5');
 	color: getColor('cold-2');
 	font-weight: 900;
-	.next-up__bottom & {
-		@include nTypeFoxtrotSize(3);
-		left: -5px;
-		top: -25px;
+	display: block;
+	margin: 0;
+
+	.next-up__header & {
+		font-size: 18px; // FIXME: revist upon type scale revisions
+		left: 10px;
+		top: -10px;
+		padding: 0 5px;
 	}
+
+	.next-up__bottom & {
+		font-size: 30px;  // FIXME: revist upon type scale revisions
+		top: -12px;
+		left: -10px;
+		padding: 0 15px;
+	}
+}
+
+.next-up__bottom__wrapper {
+	margin-top: 45px;
+	border-top: 1px solid getColor('warm-3');
+}
+
+.next_up__bottom__article_list {
+	margin: 45px 0;
+	padding-left: 20px;
+}
+
+.next_up__bottom__article {
+	list-style: none;
 }
 
 .next-up__intro__header {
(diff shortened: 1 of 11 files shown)


client/components/read-next/main.scss
```diff
@@ -38,15 +62,38 @@
...
-	&::before {
-		content: '>';
-		speak: none;
-		position: absolute;
-		top: -2px;
-		left: -15px;
+	display: inline-block;
+
+	.next-up__header & {
+		font-size: 24px;
```
i-like-robots: Is it OK to stick with the type scale (size 2)? If it doesn't match the design that should be OK, we'll be looking at the scale again soon.

client/components/read-next/main.scss
```diff
@@ -38,15 +62,38 @@
...
+			@include nextIcon(arrow-right, getColor('cold-2'), 17);
+			content: '';
+			speak: none;
+			position: absolute;
+			top: 8px;
+			left: -22px;
+		}
+
+		@include oGridRespondTo(M) {
+			font-size: 30px;
```
i-like-robots: Same as above

imranolas:
screen shot 2015-09-28 at 16 06 27
screen shot 2015-09-28 at 16 06 45

client/components/read-next/main.scss
```diff
@@ -55,6 +102,7 @@
 	color: oColorsGetColorFor(timestamp, text);
 	text-transform: uppercase;
 	white-space: nowrap;
+	display: none;
```
i-like-robots: Is there a need for the other styles in this block?

client/components/suggested-reads/main.js
```diff
@@ -3,7 +3,8 @@ var fetchres = require('fetchres');
 
 module.exports = {
 	init: function(flags) {
-		if (!flags.get('articleSuggestedRead')) { return; }
+		return;
```
i-like-robots: Remove the dependency and init call?

server/controllers/article.js
```diff
@@ -90,7 +91,14 @@ module.exports = function(req, res, next) {
 					standFirst: article[0] ? article[0].item.editorial.standFirst : ""
 				}),
 				socialMediaImage(article[1]),
-				res.locals.flags.articleSuggestedRead && article[0] ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : Promise.resolve()
+				res.locals.flags.articleSuggestedRead && article[0] ? readNext(article[0], res.locals.flags.elasticSearchItemGet) : Promise.resolve(),
+				getSuggested(article[0]).then(function(it) {
+					return api.contentLegacy({
+						uuid: (it && it.ids) || [],
+						useElasticSearch: res.locals.flags.elasticSearchItemGet,
+						type: 'Article'
```
i-like-robots: No type option

server/controllers/article.js
```diff
@@ -100,6 +108,7 @@ module.exports = function(req, res, next) {
 			var article = results[1];
 			var mainImage = results[3];
 			var readNextArticle = results[4];
+			var readNextArticles = results[5].map(function(it) { return articlePodMapping(it); });
```
i-like-robots: fat arrow all the things =)

views/partials/related/suggested-reads.html
```diff
@@ -1,12 +1,14 @@
...
-				<a href="/{{item.id}}" data-trackable="related">
-					<h3 class="article__suggested-reads__item__headline">{{item.title.title}}</h3>
-				</a>
-			</div>
-		</article>
-	{{/each}}
-</div>
+{{#if items.length}}
+	<div class="article__suggested-reads__article-list" role="group" aria-labelledBy="article__suggested-reads__title" >
+		<span id="article__suggested-reads__title" class="article__suggested-reads__title">Recommended:</span>
```
i-like-robots: Should this be a heading (h3 like the read next)?

client/components/read-next/main.scss
imranolas: Wasn't sure whether this element is expected to be in the final designs but wanted to keep the styles incase. Happy to remove completely in the meantime as we can always retrieve from history.

client/components/suggested-reads/main.js
imranolas: Will do.

client/components/read-next/main.scss
imranolas: I'm apprehensive to stray from the designs as they'll look pretty bad in this case. I'll mark them with a comment to be revisited upon type scale revisions.

Issue #927, opened 2015-09-25
Trial barrier styling is currently broken


ifyio:
cc/ @adambraimbridge @ironsidevsquincy @keirog 
Trial barrier styling is currently broken. May have something to do with the upgrade to o-grid

![image](https://cloud.githubusercontent.com/assets/675909/10103804/fc1a14ec-639d-11e5-80ef-d0ceca69e66b.png)

PR #932, opened 2015-09-28
Fix barrier grid layout issues

keirog:
I missed a bit when upgrading to o-grid 4. Also removes some redundant styling since we introduced default link styles.


keirog:
Resolves #927.

PR #937, opened 2015-10-01
Health checks

imranolas:
Added checks to verify CAPI and ES are running and accessible from next-article.


i-like-robots:
@aintgoin2goa @matthew-andrews

matthew-andrews:
seems good :+1:

PR #938, opened 2015-10-01
Social redesign

adgad:
TODO: 
- [x] Upload new share icons to social data set
- [x] Use non-branch version of o-share

share


client/components/share/main.js
```diff
@@ -1,10 +1,27 @@
...
+		fetch(`/article/${articleId}/social-counts?url=${url}`)
+			.then(fetchres.json)
+			.then(function(counts) {
+				shareCount.textContent = counts.shares + ' shares';
+			});
+	}
+
+}
 
 exports.init = function() {
```
keirog: Maybe add a check to make sure this only happens once

server/controllers/related/social-counts.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+const fetchres = require('fetchres');
+const cacheControl = require('../../utils/cache-control');
+
+function getShareCounts(articleUrl) {
+	let url = 'https://ft-next-sharedcount-api.herokuapp.com/v1/getCounts?groupby=url' +
+			'&services=facebook,gplus,twitter,stumbleupon,reddit&metrics=comments,shares,votes,endorsements&source=spoor' +
+			'&urls=' + articleUrl;
```
ironsidevsquincy: template string? could split the services into a `const`

client/components/share/main.js
```diff
@@ -1,10 +1,27 @@
...
+function loadShareCount() {
+	let shareCount = document.querySelector('.js-share-count');
+	let article = document.querySelector('.article');
+	let articleId = article.getAttribute('data-content-id');
+	if(shareCount) {
+		let url = shareCount.getAttribute('data-shared-url');
+		fetch(`/article/${articleId}/social-counts?url=${url}`)
+			.then(fetchres.json)
+			.then(function(counts) {
+				shareCount.textContent = counts.shares + ' shares';
```
ironsidevsquincy: template string

server/stylesheets/social.xsl
```diff
@@ -0,0 +1,28 @@
...
+												<a target="_blank" href="whatsapp://send?text={$encodedTitle}%20-%20https://next.ft.com/content/{$id}" data-trackable="whatsapp"><i>Whatsapp</i></a>
+										</li>
+								</ul>
+						</div>
+						<span class="article__share__count js-share-count" data-shared-url="{$webUrl}"></span>
+				</div>
+			</xsl:if>
+		</xsl:template>
+
+</xsl:stylesheet>
```
ironsidevsquincy: seems to be some double tabs in here

server/controllers/article.js
```diff
@@ -88,7 +88,11 @@ module.exports = function(req, res, next) {
 					fullWidthMainImages: res.locals.flags.fullWidthMainImages ? 1 : 0,
 					reserveSpaceForMasterImage: res.locals.flags.reserveSpaceForMasterImage ? 1 : 0,
 					suggestedRead: res.locals.flags.articleSuggestedRead ? 1 : 0,
-					standFirst: article[0] ? article[0].item.editorial.standFirst : ""
+					standFirst: article[0] ? article[0].item.editorial.standFirst : "",
+					renderSocial: res.locals.flags.articleShareButtons ? 1 : 0,
+					id: extractUuid(article[1].id),
+					webUrl: article[0] && article[0].item  && article[0].item.location ? article[0].item.location.uri : article[1].webUrl,
```
ironsidevsquincy: can we not always use one? e.g. just `article[1].webUrl`?

server/controllers/related/social-counts.js
commuterjoy: This is an api without a proper owner - it was set up during a hack week.

(Just saying, don't let that stop you using it)

server/controllers/article.js
adgad: So CAPIv2 returns a URL which isn't the same as the URLs users share (e.g. http://www.ft.com/cms/s/b219ab44-94fb-11e4-b32c-00144feabdc0.html - without the /s/0/) so have no shares.

Happy to use *only* CAPI1 for now - but this will definitely need looking at once Next hits critical mass - and I presume at that point (or some future point) `article[1].webUrl` will return next URLs.

server/controllers/related/social-counts.js
commuterjoy: could really do with a non-herokuapp end point

PR #944, opened 2015-10-05
Fix inline related story to have no image

adgad:
/cc @i-like-robots 

Before: 
screen shot 2015-10-05 at 17 18 15

After: 
screen shot 2015-10-05 at 17 18 07


i-like-robots:
:+1:

PR #946, opened 2015-10-07
B2b barrier

ifyio:
cc/ @commuterjoy @aintgoin2goa


client/components/barriers/_barrier-grid.scss
```diff
@@ -85,12 +85,42 @@
 	}
 }
 
+.barrier-grid__section--style2 {
+	border-bottom-style: none;
+	padding-bottom: 0 !important;
```
aintgoin2goa: why do you need the !important bits.  what are you overriding?

aintgoin2goa:
:+1:

PR #949, opened 2015-10-09
Add healthcheck for comments

adgad:
/cc @matthew-andrews @i-like-robots

Comments on Next have been broken for a couple of days without any alerts, so adding a (low priority) health check for SUDS.


matthew-andrews:
:+1:

PR #950, opened 2015-10-12
Pass in flag to be able to use elasticSearchOnAws

matthew-andrews:


server/lib/read-next.js
```diff
@@ -4,22 +4,25 @@ var articlePodMapping = require('../mappings/article-pod-mapping');
...
-		.then(function(storyPackageArticle) {
-			return articlePodMapping(storyPackageArticle);
-		})
-		.catch(function(error) {
-			return ;
-		});
+				uuid: parent.item.package[0].id,
+				useElasticSearch: useElasticSearch,
+				useElasticSearchOnAws: useElasticSearchOnAws
+			})
```
i-like-robots: ;

matthew-andrews: oops

matthew-andrews: i don't believe the linting is working properly!

i-like-robots: Nope! It let me merge a file with spaces instead of tags 🙀🙀

matthew-andrews: bad!

PR #955, opened 2015-10-14
Turn elastic search on in smoke tests

wheresrhys:
@matthew-andrews I have a feeling the smoke tests were all always hitting capi


matthew-andrews:
:+1:

PR #968, opened 2015-10-15
Remove comments on barrier

TheoLeanse:
/cc @ifyio @commuterjoy 
Wrapped the html responsible for displaying comments in an 'unless' helper


ifyio:
:+1:

PR #969, opened 2015-10-15
uses comments property of the viewModel

TheoLeanse:
/cc @ifyio @commuterjoy 

@andygnewman and I thought it would be clearer to depend on the `comments` property of the viewModel, which is set to `null` in the article controller whenever barriers are shown.


ifyio:
:+1:

PR #975, opened 2015-10-19
Fix background for trial barriers on IE8

TheoLeanse:
Previously displaying a fragmented background, now solid white.


i-like-robots:
Barriers are now being removed from this app https://github.com/Financial-Times/next-article/pull/1181

PR #989, opened 2015-10-21
content handling

i-like-robots:
Previously we stepped through each controller in sequence until one stuck and sent a response. This change means we figure out what the requested content is and then call the appropriate controller. This will help when podcast IDs become proper GUIDs and we create new controllers to handle V3 content.


andygnewman:
+1 Looks good

PR #990, opened 2015-10-22
appeasing the linting gods

andygnewman:
with bump of nbt to 5 we get stricter linters.

### DIFF
--- a/client/components/comments/sidenotes.js
+++ b/client/components/comments/sidenotes.js
@@ -94,7 +94,7 @@ function getUserData(){
 }
 
 function loadSideNotes(){
-	return new Promise(function(resolve, reject) {
+	return new Promise(function(resolve) {
 		Livefyre.require(['sidenotes#1', 'auth'], function (Sidenotes, Auth) {
 			resolve([Sidenotes, Auth]);
 		});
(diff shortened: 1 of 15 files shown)


client/components/comments/sidenotes.js
```diff
@@ -132,7 +132,7 @@ function addTracking(app){
 			source: 'next-article'
 		}
 	};
-	app.on('sidenotes.commentPosted', function(data){
+	app.on('sidenotes.commentPosted', function(){
```
i-like-robots: should `function(){` be `function() {` ?

client/components/comments/traditional.js
```diff
@@ -24,7 +24,7 @@ module.exports.init = function(uuid, flags) {
 		commentLink.textContent = commentCount;
 		document.querySelector('.article__share').appendChild(commentLink);
 	});
-	OComments.on('tracking.postComment', function (ev) {
+	OComments.on('tracking.postComment', function () {
```
i-like-robots: Surprised it doesn't complain about this - probably a good thing!

andygnewman: At some point we should also fix the 991 problems, feels like there should be a Jay-z reference in there...

PR #997, opened 2015-10-23
Tie article to a specific version of next-ft-api-client

matthew-andrews:


matthew-andrews:
So that we can release the `es-only` version of next-ft-api-client

PR #998, opened 2015-10-26
Elastic v3

i-like-robots:
- [x] Display title, author, main image and content
- [x] Display tags and primary tag
- [x] Display more ons
- [ ] Display suggested reads
- [x] Integrate DFP
- [x] Integrate barriers and first click free
- [x] Integrate Open Graph
- [x] Integrate Twitter card

It's probably OK to merge this without reaching full feature parity as it's behind a flag just so the PR doesn't get ridiculous.

After we drop support the V1 and V2 indexes this controller can be refactored somewhat, a lot of the mapping about will be redundant.


matthew-andrews:
:+1:

server/controllers/negotiation.js
```diff
@@ -40,6 +47,7 @@ function getArticleV1(contentId, flags) {
 function getArticleV2(contentId, flags) {
 	return api.content({
 		uuid: contentId,
+		inex: 'v2_api_v2',
```
andygnewman: was this meant to be `index: `

matthew-andrews:
:boom:

andygnewman:
:8ball:

PR #1012, opened 2015-10-28
Fix .catch in article legacy so that it actually works, fix bug that …

matthew-andrews:
…broke ‘7dae9ace-7996-11e5-933d-efcdc3c11c89’


andygnewman:
:+1:

PR #1020, opened 2015-10-29
Use FT.com as an API to rescue people who hit non-CAPI-supported-articles

matthew-andrews:
@wheresrhys 
e.g. http://next.ft.com/content/8f88c930-d00a-11da-80fb-0000779e2340

### DIFF
--- a/server/controllers/negotiation.js
+++ b/server/controllers/negotiation.js
@@ -127,7 +127,7 @@ module.exports = function negotiationController(req, res, next) {
 				return res.redirect(302, `${webUrl}${webUrl.includes('?') ? '&' : '?'}ft_site=falcon&desktop=true`);
 			}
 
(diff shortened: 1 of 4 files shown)


server/controllers/negotiation.js
```diff
@@ -66,7 +67,15 @@ function getArticleV3(contentId) {
...
-	});
+	})
+		// Some things aren't in CAPI v3 (e.g. Syndicated content)
+		.catch(function(error) {
+			if (fetchres.originatedError(error)) {
+				return;
+			} else {
+				throw error;
+			}
+		});
```
i-like-robots: Good call

matthew-andrews:
Also, this, https://app.getsentry.com/nextftcom/ft-next-article/group/90618238/

PR #1029, opened 2015-10-30
Image improvements

i-like-robots:
- Defines rules based on image size: inline (<= 350px), centre (> 350 and < 700) or full width (>= 700)
- Translates all images into figures with captions if available
- Adds placeholders to all images to avoid reflowing text when they load

## Screenshots

screen shot 2015-10-30 at 16 18 19
screen shot 2015-10-30 at 16 18 42
screen shot 2015-10-30 at 16 22 09
screen shot 2015-10-30 at 16 19 07
screen shot 2015-10-30 at 16 19 39


client/components/article/_main.scss
```diff
@@ -190,15 +169,11 @@
...
-		color: inherit;
-	}
-
-	@include print {
+	@media print {
 		&:after {
 			content: 'Print a single copy of this article for personal use. Contact us if you wish to print more to distribute to others. © The Financial Times Ltd.';
 			display: block;
-			border-top: 1px solid #000000;
+			border-top: 1px solid #000;
```
andygnewman: Isn't there an o-color equivalent for this?

client/components/article/_table.scss
```diff
@@ -19,7 +19,7 @@
 		}
 	}
 	tbody tr:nth-child(2n+1) {
-		background-color: $box-colour;
+		background-color: #eae0d7;
```
andygnewman: Any origami color for this? Just checked and there isn't suggest it either gets added or use one that exists.

client/components/article/_main.scss
i-like-robots: It's for print so black is black ¯\_(ツ)_/¯

andygnewman:
Looks good to me.

With the work that I'm doing, moving some of the items around I'm bound to get some merge conflicts, (but not too overwhelming). I'll also further simplify some of the XSLT - eg. taking social out of it and doing it as a partial.

PR #1031, opened 2015-11-02
Don't crash in read next when there is no article next

matthew-andrews:
https://app.getsentry.com/nextftcom/ft-next-article/group/90903508/


andygnewman:
:+1:  hadn't considered an article might not have an article...

PR #1034, opened 2015-11-02
Notify saucelabs does nothing

matthew-andrews:


ironsidevsquincy:
Really? This was working, what's changed?

matthew-andrews:
it just always comes up as passed

ironsidevsquincy:
Pretty sure this was working. As I'm on bug duty can take a look

ironsidevsquincy:
Fixed by https://github.com/Financial-Times/next-article/pull/1039

PR #1037, opened 2015-11-03
the --genre styles also don't seem to be used

matthew-andrews:
@andygnewman i know we want to bring this stuff back eventually but for now seems safe to delete?


andygnewman:
@matthew-andrews happy to delete these for now as we'll introduce in a slightly different way anyway.

PR #1038, opened 2015-11-03
Cleanup legacy

i-like-robots:
- Removes legacy article controller
- Cleans up any unused dependencies and test fixtures


matthew-andrews:
:+1:

PR #1042, opened 2015-11-03
Remove old, very stale sidenotes feature

matthew-andrews:
@i-like-robots


client/components/comments/_sidenotes.scss
```diff
@@ -1,114 +0,0 @@
-/*
-	I'm truly sorry for what lies beneath this comment - I needed to override the default livefyre code
```
matthew-andrews: <3

PR #1043, opened 2015-11-03
Cleanup some old transforms and utils

i-like-robots:
- Removes <ft-content> transforms
- Updates any transforms that used them
- Cleans up any unused utils


matthew-andrews:
:+1:

server/controllers/article-v3.js
```diff
@@ -10,10 +10,6 @@ const articleXsltTransform = require('../transforms/article-xslt');
 const bodyTransform = require('../transforms/body');
```
matthew-andrews: `bodyTransform` can be removed too iirc…

matthew-andrews: no just the `remove-body` one

PR #1054, opened 2015-11-05
add indexing to more-ons and their pods

andygnewman:
To give us some data on which more-ons users are clicking.

Will break some beacon reports, but they're fixable...


i-like-robots:
:+1:

PR #1057, opened 2015-11-05
Url sharing

seanmtracey:


client/components/share/main.js
```diff
@@ -19,10 +34,21 @@ function loadShareCount() {
 	}
 }
 
-exports.init = function() {
+exports.init = function(flags) {
+
 	var shareContainer = document.querySelector('[data-o-component=o-share]');
+
+	console.log(flags.flags.ftlabsurlsharing);
```
commuterjoy: no logs please

andygnewman:
Sean,

A few comments;

- we're moving away from the article-legacy model to the article-v3 - now in production.
- the social buttons are moving from the XSLT into the html (handlebars) templating - expected to go into production next week. (have a look in the genrification branch).

So suggest a rebase onto master as first step and then come and have a chat with us about how to fit this in with the direction in which the article app is moving.

seanmtracey:
Andy, ok, I'll look into that now.

client/components/share/main.js
```diff
@@ -2,6 +2,20 @@
 
 const OShare = require('o-share');
 const fetchres = require('fetchres');
+const oOverlay = require('o-overlay');
```
JakeChampion: `o-share` is stored in a capitalised reference, I would have this follow that convention also

Issue #1059, opened 2015-11-06
Missing specs


i-like-robots:
We are missing specs for the following modules that should really get some coverage:

- [x] server/controllers/article-helpers/suggested
- [ ] ~~server/controllers/article-helpers/barrier~~
- [x] server/controllers/access
- [x] server/controllers/related/*

<img width="1280" alt="screen shot 2015-11-09 at 17 34 17" src="https://cloud.githubusercontent.com/assets/271645/11059580/7124b298-8792-11e5-9889-f8bc60c1ad4e.png">

i-like-robots:
Suggested read spec https://github.com/Financial-Times/next-article/commit/186c1c3d552b8d30da0080b0ba172d2de6070d10

i-like-robots:
Access spec https://github.com/Financial-Times/next-article/pull/1102

PR #1062, opened 2015-11-09
fix dedupe on more-ons

andygnewman:


i-like-robots:
CC https://github.com/Financial-Times/next-article/issues/1059

PR #1068, opened 2015-11-11
[WIP] increase breakpoint for long promo and apply exapander to short

andygnewman:
- Increase breakpoint for full width promo-boxes to either 100 words plus image or 150 words no image.
- Make the expander only kick in if more than 3 paragraphs.
- Apply expander to logic to inline promo-boxes.

### DIFF
--- a/server/stylesheets/promo-box.xsl
+++ b/server/stylesheets/promo-box.xsl
@@ -2,9 +2,9 @@
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
 
     <xsl:template match="/html/body/promo-box">
-      <xsl:variable name="longBoxWordBoundaryImage" select="45" />
-      <xsl:variable name="longBoxWordBoundaryNoImage" select="90" />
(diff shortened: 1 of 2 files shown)


server/stylesheets/promo-box.xsl
```diff
@@ -35,12 +35,26 @@
...
         <xsl:otherwise>
-          <aside class="promo-box ng-inline-element" data-trackable="promobox" role="complementary">
-            <div class="promo-box__wrapper">
-              <xsl:apply-templates select="current()" mode="default-title" />
-              <xsl:apply-templates />
-            </div>
-          </aside>
+          <xsl:choose>
+            <xsl:when test="$contentParagraphs > $expanderParaBreakPoint">
+              <aside class="promo-box ng-inline-element o-expander" data-trackable="promobox" role="complementary" data-o-component="o-expander" data-o-expander-shrink-to="0" data-o-expander-count-selector=".promo-box__content__extension">
```
i-like-robots: We're duplicating a lot of the markup, we can probably DRY this up

PR #1073, opened 2015-11-12
Add a healthcheck for elevated error rate

adgad:
/cc @i-like-robots @andygnewman @phamann @commuterjoy 

Adds a health check for article that alerts when the rate of errors in the last 10 minutes is more than 3 times the average over the last week.


phamann:
:+1:

PR #1074, opened 2015-11-12
add a print button

andygnewman:
- Puts a print button on the article 'toolbar'
- When pressed brings up the print dialog box
- Not displayed on default or small screen size

Desktop
screen shot 2015-11-12 at 15 29 17

Small Screen (doesn't appear)
screen shot 2015-11-12 at 15 30 03

### DIFF
--- a/client/components/share/main.scss
+++ b/client/components/share/main.scss
@@ -39,6 +39,20 @@
 		}
 	}
 
+	&__print {
+		display: none;
+		border-bottom-width: 0;
+		float: left;
+		margin-right: 16px;
+		&:before {
+			@include nextIcon(printer, getColor('cold-1'), 25);
+			content: '';
+		}
+		@include oGridRespondTo(M) {
+			display: block;
+		}
+	}
+
 	.o-share__action a {
 		border-bottom-width: 0;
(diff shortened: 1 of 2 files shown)


views/partials/social.html
```diff
@@ -20,6 +20,7 @@
   {{/if}}
   {{#if @root.flags.saveForLater}}
     <div class="article__save">
+      <a class="article__share__print" href="javascript:window.print()" data-trackable="print"></a>
```
i-like-robots: Best to have some text here `<span class="n-util-visually-hidden">Print this page</pan>`

client/components/share/main.scss
```diff
@@ -39,6 +39,22 @@
...
+	&__print {
+		display: none;
+		border-bottom-width: 0;
+		float: left;
+		margin-right: 16px;
+		&:before {
+			@include nextIcon(printer, getColor('cold-1'), 25);
+			content: '';
+			position: relative;
+			top: 0;
```
i-like-robots: Are the `position` and `top` properties needed?

i-like-robots:
:ship:

PR #1081, opened 2015-11-13
Make the graphite check even less thresholdy and less severe

adgad:
/cc @wheresrhys @commuterjoy @phamann @andygnewman


phamann:
:+1:

PR #1089, opened 2015-11-17
Add generic share received message

JakeChampion:
If a user came to an article because somebody shared it with them with a share token, we want to display a generic message to mention this to the user.

### DIFF
--- a/server/controllers/article.js
+++ b/server/controllers/article.js
@@ -144,6 +144,7 @@ module.exports = function articleV3Controller(req, res, next, payload) {
 
 	if (req.get('FT-Labs-Gift') === 'GRANTED') {
 		payload.shared = true;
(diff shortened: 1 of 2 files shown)


andygnewman:
:+1:

commuterjoy:
A word of warning about this.

If we don't vary the contents of the page by `ft-labs-gift` this will cache in the CDN with this message. It's only because the request comes with a querystring (and thus forms a unique cache key, separate to the normal article URLs) that this PR is ok to ship.

But I am wondering whether adding `ft-labs-gift` to the outbound vary header would prevent future confusion.

JakeChampion:
@commuterjoy I thought we were varying by the header already, if we are not, I think we should be.

PR #1097, opened 2015-11-19
myFtApi flag is the new userPreferencesAPI flag

keirog:


keirog:
Got IRL +1 from @laurieboyes

PR #1098, opened 2015-11-19
Adjusted code to handle more than one o-trigger

seanmtracey:
Adjusted sharing component to bind events to multiple triggers.


andygnewman:
OK to squerge

PR #1103, opened 2015-11-19
related-boxes (the new promo-boxes)

andygnewman:
TO DO
- [x] XSLT - identify related items that are articles with uuids that we can fetch later
- [x] XSLT - strip any existing markup from articles we will fetch later.
- [x] XSLT - transform for presentation any other `<ft-related>` items without amending content except for adding links to image / headline and a Read More if `<ft-related>` has a url attribute.
- [x] Cheerio - Fetch the article(s) and create the html within the body
- [x] Cheerio - Ensure it works for an article without an image
- [x] Styling - bring into line with promo-box styling - can amend later once live

### DIFF
--- a/client/components/article/_related-box.scss
+++ b/client/components/article/_related-box.scss
@@ -1,0 +1,64 @@
+.related-box {
+	&:after {
+		content: '';
+		display: table;
+		clear: both;
+	}
+}
+
+.related-box__title {
+	margin-top: -48px;
+}
+
+.related-box__title__name {
+	@include oTypographySerifDisplay(m);
+	display: inline-block;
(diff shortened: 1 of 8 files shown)


client/components/article/_related-box.scss
```diff
@@ -0,0 +1,62 @@
...
+		clear: both;
+	}
+}
+
+.related-box__title {
+	margin-top: -48px;
+}
+
+.related-box__title__name {
+	@include nTypeFoxtrot(3);
```
matthew-andrews: can we start switching to `o-typography` mixins for new code?

PR #1110, opened 2015-11-23
get all the share counts

andygnewman:
Add up all the shares across all different url configurations
Add in linkedin shares


server/controllers/related/social-counts.js
```diff
@@ -1,35 +1,47 @@
...
-	let url = `https://ft-next-sharedcount-api.herokuapp.com/v1/getCounts?groupby=url` +
-			`&services=${services}&metrics=${metrics}&source=next-article` +
-			`&urls=${articleUrl}`;
-	return fetch(url, { timeout: 3000 })
-		.then(fetchres.json);
+function getWebUrls (articleId) {
+	return [
+		`http://www.ft.com/cms/s/${articleId}.html`,
+		`http://www.ft.com/cms/s/0/${articleId}.html`,
+		`http://www.ft.com/cms/s/2/${articleId}.html`
```
commuterjoy: Is generating the correct URL via API v1 simple? There should only be one. 

Plus you will miss URLs like this,

http://api.ft.com/content/items/v1/0166c361-d62d-33eb-b2de-a2a5316d34e5

yields,

http://ftalphaville.ft.com/2015/08/06/2136475/corbyns-peoples-qe-could-actually-be-a-decent-idea/

Also, '3' is a valid code, if you are going to do it the above way ...

http://www.ft.com/cms/s/3/7893be18-668e-11e5-a57f-21b88f7d973f.html

PR #1117, opened 2015-11-27
More ons dedupe

andygnewman:
Objective:
Not repeat stories from preceding more-ons in later more-ons.

This retains each more-on as a separate call with a response that can be mapped back on to the appropriate element on the page.
Subsequent more-on calls therefore repeat the processing of preceding more-ons to know which articles to dedupe against.

The functionality has a built in limitation that will not process anymore than 5 more-ons as processing.
It also limits the number of articles that can be requested to 10 per more-on.

Plus it has a full test suite.


matthew-andrews:
:+1: (and an extra one for the tests)

PR #1122, opened 2015-11-30
test for social counts

andygnewman:


andygnewman:
@i-like-robots Test is failing - TypeError: Cannot read property 'then' of undefined
      at Context.<anonymous> (test/server/controllers/related/social-counts.test.js:43:34) I suspect it's promise related, in my 'nock' ing.

PR #1129, opened 2015-12-01
handle thin images better

andygnewman:
Some images are thin (width < 150px). Currently these are treated as inline images and when in M screen size and below, they are centred full width.

eg.  on iPhone 5
screen shot 2015-12-01 at 14 42 24

screen shot 2015-12-01 at 14 41 57

This PR introduces a new image variant of 'thin' for images with width < 150px and keeps them inline regardless of screen width.

eg. on iPhone 5

screen shot 2015-12-01 at 14 33 33

screen shot 2015-12-01 at 14 34 00

It also proposed to change the breakpoint at which inline images are full width centred, from M to S.

Eg on a Nexus 7 (portrait)

Current

screen shot 2015-12-01 at 14 39 50

Proposed

screen shot 2015-12-01 at 14 40 06

### DIFF
--- a/client/components/article/_images.scss
+++ b/client/components/article/_images.scss
@@ -10,13 +10,19 @@
 
 .article-image--inline {
 
-	@include oGridRespondTo(M) {
+	@include oGridRespondTo(S) {
 		float: left;
 		clear: left;
 		max-width: 100%;
 		margin-right: 1em;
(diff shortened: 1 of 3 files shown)


matthew-andrews:
very neat :+1:

PR #1131, opened 2015-12-02
fix issue with slideshow selector

andygnewman:
Selector for slideshow was a bit greedy.


matthew-andrews:
:+1:

matthew-andrews:
is it possible for a test for dilip's case?

andygnewman:
Was thinking of pulling the image / toc extraction logic out into a separate transform which I can put tests against. Will put that on my agenda for today.

PR #1132, opened 2015-12-02
Fix unclickable links on bb7

adgad:
/cc @i-like-robots @andygnewman @matthew-andrews 
Currently all article links (inc. share, save, in-article links) are not clickable on BB7 devices. Adding a clearfix fixes that ¯\_(ツ)_/¯


matthew-andrews:
hmm… if we fix this in `n-layout` would it fix it globally?  also gonna need a comment ……

`{{! handlebars comment probably best so it doesn't get into the output}}`

matthew-andrews:
also, any idea *why*?

adgad:
Yeah hold off for now...I'm still digging into the rabbit hole

matthew-andrews:
:rabbit2:

adgad:
Traced it down to this: https://github.com/Financial-Times/o-grid/pull/116

PR #1133, opened 2015-12-02
tests and rewrite for extracting main image and toc

andygnewman:
@matthew-andrews 

The tests we discussed earlier - writing the tests enabled me to rewrite it into something a bit simpler and hopefully more readable.


matthew-andrews:
lovely :+1:

PR #1137, opened 2015-12-03
use correct article title in sharing

andygnewman:
article title in share template was out of date vs what's returned by v3


matthew-andrews:
does it need to be `encodeURIComponent`'d?

matthew-andrews:
i think we should have a helper for that if so ^^

andygnewman:
You're right, does need encoding, so sorting that out now.

matthew-andrews:
ahh, sorry…  i think you can do this more simply with `{{#encode title}}` in the template ^^
https://github.com/Financial-Times/n-handlebars/blob/master/src/helpers/encode.js

matthew-andrews:
or maybe it's `{{#encode}}{{title}}{{/encode}}`

andygnewman:
Good point, it's actually {{encode title mode='uriComponent'}}

PR #1140, opened 2015-12-03
improve print stylesheet

andygnewman:
To Do
- [x] enable images to be printed
- [x] don't page break within images
- [x] firefox page breaking and don't stop part way through an article
- [x] enable promoboxes to be printed
- [x] fix date display


commuterjoy:
i think you should go down to PC World with @dilipvaghela and test this :)

PR #1142, opened 2015-12-04
follow button beside primary tag behind flag

andygnewman:
put follow button beside primary tag.
put behind a flag so can test in logged in environment.


matthew-andrews:
:+1:

PR #1145, opened 2015-12-04
Added URLs to social sharing links for labs-o-share

seanmtracey:
Added the URLs to the social sharing icons template for labs-o-share.

### DIFF
--- a/views/partials/labs-share.html
+++ b/views/partials/labs-share.html
@@ -8,16 +8,16 @@
 				<h3>Share on social networks:</h3>
 				<ul>
 					<li class="labs-o-share__action labs-o-share__action--twitter">
-						<a data-trackable="twitter" href="https://twitter.com/intent/tweet?url=https://next.ft.com/content/{{id}}&amp;text={{og.title}}&amp;via=FT"><i>Twitter</i></a>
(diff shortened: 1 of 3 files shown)


views/partials/labs-share.html
```diff
@@ -8,16 +8,16 @@
 				<h3>Share on social networks:</h3>
 				<ul>
 					<li class="labs-o-share__action labs-o-share__action--twitter">
-						<a data-trackable="twitter"><i>Twitter</i></a>
+						<a data-trackable="twitter" href="https://twitter.com/intent/tweet?url=https://next.ft.com/content/{{id}}&amp;text={{og.title}}&amp;via=FT"><i>Twitter</i></a>
```
andygnewman: Would suggest you encode the title as special chars will lead to failures, eg.

{{encode og.title mode='uriComponent'}}

encode is a n-handlebars helper

seanmtracey: Ok, Is n-handlebars already a dependancy in next-article? I can't seem to spot it in the package/bower.json, but when I edit the template, the helper is being used.

andygnewman:
@seanmtracey presume we're OK to close this one without merging?

PR #1150, opened 2015-12-07
Quickly hack in a ?fragment view for articles

matthew-andrews:


andygnewman:
Had come to a similar conclusion for my hack that I would need an article endpoint that could render just the body.

PR #1151, opened 2015-12-07
use ?fragments to return just the article body

andygnewman:
@matthew-andrews 

?fragments will just return the body of the article

### DIFF
--- a/server/controllers/article.js
+++ b/server/controllers/article.js
@@ -74,7 +74,6 @@ function getMoreOnTags(primaryTheme, primarySection) {
 
 module.exports = function articleV3Controller(req, res, next, payload) {
 
-
 	let asyncWorkToDo = [];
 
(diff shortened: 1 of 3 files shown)


server/controllers/article.js
```diff
@@ -156,10 +156,11 @@ module.exports = function articleV3Controller(req, res, next, payload) {
 		.then(() => {
 			if (req.query.fragment) {
 				payload.layout = 'vanilla';
+				return res.set(cacheControlUtil).render('./partials/body', payload);
```
matthew-andrews: can probably move the `res.set(cacheControlUtil);` to before the `if` … 

Also do we need to `return`?

matthew-andrews:
one small comment otherwise :+1:

matthew-andrews:
:+1:

PR #1154, opened 2015-12-08
disable nightwatch and minor print layout change

andygnewman:
cc: @matthew-andrews


matthew-andrews:
:+1: i think myft really ought to be dealing with its own print styles :-/

PR #1157, opened 2015-12-08
add bare fragment and wrap in article--fragment class

andygnewman:
Enable a fragment without vanilla layout (pure html, no css)
Wrap the fragment in a div with class article__fragment so can size the iFrame


matthew-andrews:
:+1:

PR #1158, opened 2015-12-08
Ads 598

leggsimon:
Puts in an ad leaderboard within article to take advantage of `article--brand` header styling


andygnewman:
:+1:  if the ad is always going to be the same height there it would be useful to add that height to the placeholder to stop the page from reflowing when the ad appears. Doesn't have to be done now, but a thought.

leggsimon:
@andygnewman that does already happen here https://github.com/Financial-Times/next-ads-component/pull/94/files#diff-29e6b718718c515120c08fdf3752e741R5

PR #1161, opened 2015-12-09
Improve accessibility with aria-labelledby

alicebartlett:
From the W3C-ARIA Spec [http://www.w3.org/TR/wai-aria/roles#region]:

> Authors SHOULD ensure that a region has a heading referenced by aria-labelledby.
This heading is provided by an instance of the standard host language heading
element or an instance of an element with role heading that contains the heading
text.

In reality what this does is means that when the whole region is navigated to
using a screen reader the screen reader will read out the heading and
-importantly- allow the user to skip the whole region if they want to. Without
the aria-labelledby this doesn't work.

Here's a nice GIF of voiceover in safari showing the behaviour after this change:
aria-labelledby1

### DIFF
--- a/views/partials/more-ons.html
+++ b/views/partials/more-ons.html
@@ -2,12 +2,12 @@
 	<div class="js-story-package" data-trackable="story-package"></div>
 	{{#if moreOns}}
 		{{#each moreOns}}
-			<div data-trackable="more-on-{{@index}}" role="region">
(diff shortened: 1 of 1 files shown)


andygnewman:
:+1:

i-like-robots:
:ship:

andygnewman:
Already incorporated in other PR's

PR #1166, opened 2015-12-14
remove hrs transform

andygnewman:
Following on from this transform being migrated to next-es-interface https://github.com/Financial-Times/next-es-interface/pull/229


commuterjoy:
thanks :)

+1

PR #1169, opened 2015-12-15
ensure primary tag follow doesnt overlap tag

andygnewman:
To stop follow button from overlapping primary tag on mobile display.


matthew-andrews:
let me take this fo ra spin

matthew-andrews:
screen shot 2015-12-15 at 10 08 36

LGTM!

andygnewman:
Is that the longest topic we have?

PR #1171, opened 2015-12-16
Related article styling

andygnewman:
Differentiate the styling of related boxes that are specifically articles vs those we don't know their type.

From this;
screen shot 2015-12-16 at 17 22 37

To this;
screen shot 2015-12-16 at 17 22 13


matthew-andrews:
really great to see this !

PR #1172, opened 2015-12-18
handle a related article that aready has a title field

andygnewman:
Wasn't expecting that, but as we have them, we should retain existing title and put it in the right place.


matthew-andrews:
:+1:

PR #1177, opened 2015-12-18
Stop diluting our Google juice by including ‘rel=canonical’s #SEO

matthew-andrews:
screen shot 2015-12-18 at 14 31 23

### DIFF
--- a/views/article.html
+++ b/views/article.html
@@ -5,6 +5,7 @@
 	{{#if dfp.dfpZone}}
 	<meta name="dfp_zone" content="{{dfp.dfpZone}}" />
 	{{/if}}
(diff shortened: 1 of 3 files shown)


commuterjoy:
+1

... though Google doesn't index us atm?

matthew-andrews:
I don't think it indexes us but it would be able to reach Next stuff if it discovers and tries to follow an old `m.ft.com` URL.

commuterjoy:
yup!

matthew-andrews:
still then, without clearing cookies, it won't be able to get back to that `canonical` URL…  not sure what the consequence of that is… :speak_no_evil:

PR #1181, opened 2015-12-21
Sans barriers

commuterjoy:
- [x] Remove article code from `next-article`
- [x] Move barriers to `next-product-selector`
- [x] Think of less bad name than product selector - Eg, ft.com/packages
- [x] Deprecate `next-barriers` middleware
- [x] Route barriers to product selector in fastly
- [x] Carry the article title through to some barriers
- [x] Ensure `vary` headers & `stale-if-error` are retained in article and product-selector
- [x] Scale up product selector app to a few more dynos


i-like-robots:
@commuterjoy Want me to pick this up and push this de-barrier-ification through?

commuterjoy:
Two more things - 

- [ ] Ship https://github.com/Financial-Times/next-cdn/pull/15
- [ ] Modify https://github.com/Financial-Times/next-product-selector to render corporate barriers

server/app.js
```diff
@@ -25,19 +24,12 @@ app.post('^/preview$', require('./controllers/preview'));
 
 const uuid = '[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}';
 
-// COMPLEX: Put access middleware before barrier middleware so that access can be cached by membership
-app.use(`^/content/:id(${uuid})$`, require('./controllers/access'));
```
commuterjoy: @i-like-robots is this deletion a mistake?

i-like-robots: Previous commit removed this too https://github.com/Financial-Times/next-article/commit/afa6f7227871e3984f79025ea76d45b7b722c1ba

i-like-robots: It smells like a mistake because the controller itself and related test etc. haven't been removed.

commuterjoy: so if the article app is still used by access then I should keep this line

i-like-robots: I would think so!

PR #1185, opened 2015-12-23
single view for all content types

andygnewman:
Objective - have a single overarching 'content' template that covers podcasts and articles with more stuff in partials

Incorporate main images and more-ons for podcasts.

- [x] podcast main image (getting correct dimensions in a PR for next-es-interface https://github.com/Financial-Times/next-es-interface/pull/317)
- [x] align with changes to article (eg. dfp stuff)
- [x] check classes all aligned with n-content-body
- [x] single template
- [x] refactor coherent elements into separate partials
- [x] enable podcasts to use standard 'more on' code

Example of podcasts with new view

Podcast with a square image
screen shot 2016-03-31 at 17 00 04

Podcast with a 16:9 image
screen shot 2016-03-31 at 17 00 31


server/controllers/article-helpers/get-more-on-tags.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+const addTagTitlePrefix = require('./tag-title-prefix');
+
+module.exports = function (primaryTheme, primarySection, primaryBrand) {
+	let moreOnTags = [];
```
i-like-robots: const

server/controllers/article-helpers/podcast-main-image.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const figureVariant = require('../../transforms/figure-variant');
+const externalImagesEncoding = require('../../transforms/external-images-encoding');
+const cheerio = require('cheerio');
+
+module.exports = (mainImage) => {
+	// HACK to stop square images being too being
+	const squareImageDimensions = 394 // height of a 700 width image with 16:9 aspect ratio
+	if (mainImage.width === mainImage.height && mainImage.width > 394) {
```
i-like-robots: Is this still necessary with changes made to ingest?

server/controllers/podcast.js
```diff
@@ -26,16 +29,46 @@ module.exports = function podcastLegacyController(req, res, next, payload) {
...
 	}
 
-	asyncWorkToDo.push(
-		suggestedHelper(payload.id, [], payload.primaryTag).then(
-			articles => payload.relatedContent = articles
-		)
-	);
+	// HACK set primaryBrand to Podcasts so will populate second more on
+	payload.primaryBrand = payload.tags
+		.find(tag => tag.id === 'NjI2MWZlMTEtMTE2NS00ZmI0LWFkMzMtNDhiYjA3YjcxYzIy-U2VjdGlvbnM=');
```
i-like-robots: Shall we move this upstream? Might be tricky as podcasts still talk the old metadata language https://github.com/Financial-Times/next-es-interface/blob/master/server/lib/podcast-metadata.js

views/partials/podcast-media.html
```diff
@@ -0,0 +1,12 @@
...
+		{{#with media}}
+		<audio controls>
+			<source src="{{url}}" type="{{mediaType}}" />
+			<p>
+				Your browser does not support playing this file but you can still <a href="{{url}}">download the MP3 file</a> to play locally.
+			</p>
+		</audio>
+		{{/with}}
+	</div>
+{{/if}}
```
i-like-robots: I guess this doesn't need to be a partial... checking if the media exists is also a bit redundant, it would be better to error so we know about it!

views/partials/related/more-ons-tags.html
```diff
@@ -11,7 +11,7 @@ <h3 class="more-ons-tags__title-text">
 		</h3>
 	</div>
 	<div class="n-content-tags__intro">
-		<p>Follow the topics mentioned in this article</p>
+		<p>Follow the topics mentioned in this {{contentType}}</p>
```
i-like-robots: Hmm, this isn't true for podcasts at all as the tags are based on the show not an episode

server/controllers/article-helpers/podcast-main-image.js
andygnewman: Yes, it's an opinionated hack to not have massively square images, so whilst the ingester gets the real dimensions, it's only in the article implementation we make it more acceptable for the article rendering.

i-like-robots: Do we have similar logic elsewhere? I.E. `ratio > 0.75 && ratio < 1.33333`

i-like-robots: Or even `ratio > 0.75` for both squarish and portrait

server/controllers/podcast.js
andygnewman: I think that the next step would be to DRY out the article and podcast controllers given repetition, where some of this will be more obvious.

i-like-robots: Yep, I imagine we can use the same controller, just add a flag - `isArticle`, `isPodcast` etc

i-like-robots:
:ship:

PR #1186, opened 2015-12-24
add copyright notice to figcaption

andygnewman:
Put the image copyright notice into the figcaption, appended to longdesc when that is present.
Swap from deprecated font styling to oTypography

### DIFF
--- a/test/server/stylesheets/external-image.test.js
+++ b/test/server/stylesheets/external-image.test.js
@@ -321,6 +321,54 @@ describe('External images', () => {
 				});
 		});
 
+		it('adds caption when the copyright is present', () => {
+			return transform(
+				'<html>' +
+					'<body>' +
+						'<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>' +
+						'<p><img src="http://my-image/image.jpg" data-copyright="© Bloomberg" width="800" height="600" /></p>' +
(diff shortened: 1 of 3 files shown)


test/server/stylesheets/external-image.test.js
```diff
@@ -321,6 +321,54 @@ describe('External images', () => {
...
+			return transform(
+				'<html>' +
+					'<body>' +
+						'<p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>' +
+						'<p><img src="http://my-image/image.jpg" longdesc="This is a long description" data-copyright="© Bloomberg" width="800" height="600" /></p>' +
+						'<p>Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>' +
+					'</body>' +
+				'</html>\n'
+			)
+			.then((transformedXml) => {
```
matthew-andrews: It really makes no difference at all but did you know that the brackets are optional in fat arrow functions when you only have one variable?
`.then(variable => {`

matthew-andrews:
:+1: :+1: (one for the change and one for the baby step toward origamification)

test/server/stylesheets/external-image.test.js
matthew-andrews: I can understand if you would want to keep the brackets for single variable fat arrows for consistency though.

andygnewman: Yes I did know that, but was keeping consistency as you said.

matthew-andrews: :+1: quite a nice idea. Keeps diffs cleaner. I might start doing the same.

PR #1189, opened 2015-12-24
handle nothing to dedupe

andygnewman:
To solve these issues in sentry

https://app.getsentry.com/nextftcom/ft-next-article/issues/99106206/
https://app.getsentry.com/nextftcom/ft-next-article/issues/99107120/

### DIFF
--- a/server/controllers/related/more-on.js
+++ b/server/controllers/related/more-on.js
@@ -55,6 +55,9 @@ module.exports = function (req, res, next) {
 	let precedingMoreOnIds = [];
 
 	let dedupe = function(articlesToDedupe) {
+		if (!articlesToDedupe) {
+			return [];
+		}
 		return articlesToDedupe
 			.filter(article => isNotADupe(article.id))
 			.slice(0, count);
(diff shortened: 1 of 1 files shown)


server/controllers/related/more-on.js
```diff
@@ -55,6 +55,9 @@ module.exports = function (req, res, next) {
 	let precedingMoreOnIds = [];
 
 	let dedupe = function(articlesToDedupe) {
+		if (!articlesToDedupe) {
+			return [];
+		}
 		return articlesToDedupe
```
i-like-robots: Neater to use a ternary

server/controllers/related/more-on.js
```diff
@@ -55,6 +55,9 @@ module.exports = function (req, res, next) {
 	let precedingMoreOnIds = [];
 
 	let dedupe = function(articlesToDedupe) {
```
i-like-robots: function expressions should be declared with const (or just use a named function)

i-like-robots:
nitpicking =]

PR #1193, opened 2016-01-05
Update chrome and firefox versions for nightwatch

andygnewman:


ironsidevsquincy:
just `chrome` should give you the latest version (well, as long as we keep https://github.com/Financial-Times/next-build-tools/blob/master/config/nightwatch.json#L28 updated)

andygnewman:
Yes, just spotted that - saw your change and thought 'what idiot has left off the version number', then actually realised it was rather clever to allow it to be updated in a single place for the latest version...bravo

PR #1206, opened 2016-01-12
Allow a users followed topics to be displayed via a query string

phamann:
As discussed in person, this PR allows for a users followed myFT topics to be passed in via a query string `?myftTopics`. This will ensure that their given topics are always displayed in the tags array in top right hand corner of layout. 

#### N.b.
This is to fix an issue whereby users are getting confused as they are displayed the content in their feed for a given topic which is then not displayed on the content page.

#### Future
I have discussed with the POs and designers that we should re-think how followed topics are presented to a user when on a content page.

### DIFF
--- a/server/controllers/article-helpers/decorate-metadata.js
+++ b/server/controllers/article-helpers/decorate-metadata.js
@@ -46,7 +46,7 @@ function selectTagsMyftTagsForDisplay(article) {
 
 	return article.metadata
 		.filter(isPrimaryTag(article))
-		.filter(tag => myftTopics.some(id => id === tag.id ));
+		.filter(tag => myftTopics.some(id => id === tag.id));
 }
 
 function selectTagsForDisplay(article) {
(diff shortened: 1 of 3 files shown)


server/controllers/article-helpers/decorate-metadata.js
```diff
@@ -37,15 +37,27 @@ function selectPrimaryTag(article) {
...
+
+function selectTagsMyftTagsForDisplay(article) {
+	let myftTopics = article.myftTopics || [];
+
+	return article.metadata
+		.filter(isPrimaryTag(article))
+		.filter(tag => myftTopics.some(id => id === tag.id));
+}
+
 function selectTagsForDisplay(article) {
```
i-like-robots: Is this ever invoked?

phamann: [Yes here](https://github.com/Financial-Times/next-article/pull/1206/files#diff-9a96daee922c1a589ae9eefa5a933c4fR69)

PR #1212, opened 2016-01-18
stop promobox text overflowing when printed

TheoLeanse:
In response to Jira issue NFT-243 "Text overflowing out of promo boxes on firefox and IE, when printing articles."  

@andygnewman


andygnewman:
LGTM

PR #1213, opened 2016-01-19
Homepage follow journey

adgad:
/cc @ironsidevsquincy @andygnewman @Financial-Times/next-myft 

For a user being recommended a topic to follow on the front page and coming to an article, we want to promote that topic to them to encourage them to follow it. This may or may not be the same as the primary topic of the page.

We will give them a single large follow CTA (hiding the follow button next to the primary tag on desktop).

Also in this PR is some BEM tidy up, and changing the myFT text logo at the bottom of the page to the myFT logo.

screen shot 2016-01-21 at 12 29 38
screen shot 2016-01-21 at 12 29 46

### DIFF
--- a/client/components/_common.scss
+++ b/client/components/_common.scss
@@ -160,7 +160,8 @@
 	text-align: center;
 }
 
-.u-hide--L {
+//Want this to be capital L for consistency with o-grid definition
+.u-hide--L { 	// scss-lint:disable SelectorFormat
 	@include oGridRespondTo(L) {
 		display: none;
 	}
--- a/client/components/_common.scss
+++ b/client/components/_common.scss
@@ -15,34 +15,6 @@
 	padding: 16px 0;
 	width: 100%;
 
-	&--inline {
-		float: left;
-		@include oGridRespondTo(S) {
-			max-width: 300px;
-			margin-right: 24px;
-		}
-	}
-
-	&__title {
-		margin: -40px 0 8px 10px;
-	}
-
-	&__title-text {
-		@include oTypographySerifDisplay(m);
-		color: getColor('cold-1');
-		display: inline-block;
-		margin: 0;
-		padding: 10px;
-		padding-bottom: 0;
-		font-weight: 900;
-		background-color: inherit;
-		a {
-			color: getColor('cold-1');
-			border-bottom: 0;
-			text-decoration: none;
-		}
-	}
-
 	.o-expander__toggle {
 		@include oTypographySans(m);
 		margin-top: 12px;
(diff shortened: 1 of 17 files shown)


client/components/article/_main.scss
```diff
@@ -2,7 +2,6 @@
 .article__header {
 	position: relative;
 	background: getColor('pink');
-	font-weight: 100;
```
adgad: This didn't look like it was being used by anything

client/components/_common.scss
```diff
@@ -70,7 +71,7 @@
 	@include nLinksTopic;
 	font-style: normal;
 
-	&__medium {
+	&--medium {
```
i-like-robots: Not your bad but writing selectors this way makes them impossible to find :no_mouth:

views/article.html
```diff
@@ -22,14 +22,20 @@
...
 								{{#if primaryTag}}
 									<div class="article__primary-theme" data-trackable="primary-theme">
-										<a href="/stream/{{primaryTag.taxonomy}}Id/{{primaryTag.id}}" class="c-topic c-topic__medium c-topic--with-follow c-topic--link" data-trackable="section-link" aria-label="posted in category {{primaryTag.name}}" data-concept-id="{{primaryTag.id}}">{{primaryTag.name}}</a>
+										<a href="/stream/{{primaryTag.taxonomy}}Id/{{primaryTag.id}}" class="c-topic c-topic--medium c-topic--with-follow c-topic--link" data-trackable="section-link" aria-label="posted in category {{primaryTag.name}}" data-concept-id="{{primaryTag.id}}">{{primaryTag.name}}</a>
 										{{#ifAll @root.flags.myFtApi @root.flags.articleFollowPrimayTag}}
-											{{>next-myft-ui/templates/follow version='3' conceptId=primaryTag.id name=primaryTag.prefLabel taxonomy=primaryTag.taxonomy}}
+										{{#if tagToFollow}}
+											<span class="u-hide--L">
+										{{else}}
+											<span>
```
i-like-robots:
```
<span class="{{#if tagToFollow}}u-hide--L{{/if}}">
```

wheresrhys:
Re the logo @pauloneillft wrote some guidelines about use of *my*FT vs the logo, leaning towards using *my*FT in most instances if I remember correctly. 

I personally think we *should* aim to use the logo everywhere: given that we a) have to add `<i>my</i>FT` markup to do the italic anyway we may as well go the whole hog and do `<span class="myft-ui__logo">myFT</span>`, b) we now have fonts with reliable line heights via o-typography, so we'd only need to write ~5 variants 

But whether we do that or not, consistency should be maintained

@keirog

client/components/_common.scss
adgad: Good point. I shall unnest them all (in this file) while i'm here

adgad:
Instruction to use the logo here came from @pauloneillft - I presume it counts as a header here rather than inline text (the guidelines imply that the italicised version is for inline text).

PR #1215, opened 2016-01-21
Add basic spec for slideshow controller

i-like-robots:
We're getting closer to full coverage of the server code =]

This also picked up a bug with our current error handling (not using fetchres module to parse the response meant our check for a fetchres error was futile and we'd deliver a 50x not a 404 😳)


matthew-andrews:
\o/

PR #1218, opened 2016-01-26
incorporate bare in fragment

andygnewman:
Query String
- bare applied the vanilla layout (without headers and footers)
- this was differentiated from fragment (article body) when testing iframe vs dom insertion for open article in stream

Now decision made to go with dom insertion, the vanilla layout can be applied to fragment.

### DIFF
--- a/server/controllers/article.js
+++ b/server/controllers/article.js
@@ -150,7 +150,6 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		.then(() => {
 			res.set(cacheControlUtil);
 			if (req.query.fragment) {
-				content.layout = 'vanilla';
 				res.render('fragment', content);
 			} else {
 				content.layout = 'wrapper';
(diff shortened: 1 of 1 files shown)


server/controllers/article.js
```diff
@@ -150,9 +150,7 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		.then(() => {
 			res.set(cacheControlUtil);
 			if (req.query.fragment) {
-				if (!req.query.bare) {
-					content.layout = 'vanilla';
-				}
+				content.layout = 'vanilla';
```
i-like-robots: Vanilla will return a full document. If we just want the fragment maybe this should be `null`

commuterjoy: Also possible to specify a local layout template too.

https://github.com/Financial-Times/next-product-selector/pull/43/files

i-like-robots:
:+1:

PR #1219, opened 2016-01-27
Opt into inline follow notifications

keirog:
This will only have an effect when the *my*FT follow-email feature is
in place.


keirog:
Closing this as we're parking inline button notifications for now :tada:

PR #1229, opened 2016-02-03
Remove old ES checks

adgad:
/cc @matthew-andrews @i-like-robots @phamann 

(master build currently broken since these methods don't exist)


i-like-robots:
+1

PR #1233, opened 2016-02-08
Try out a failing master build theory

matthew-andrews:
If this is true it should probably be fixed in nbt too……


matthew-andrews:
[headers should be case insensitive]

matthew-andrews:
nope

PR #1234, opened 2016-02-09
MVT Number of More On Article Links

andygnewman:


client/components/onward-journey/main.js
```diff
@@ -28,7 +28,20 @@ function createPromise(el, url) {
...
 
-module.exports.init = function() {
+module.exports.init = flags => {
+	const moreOnNumberState = flags.get('articleMoreOnNumber');
+	let moreOnNumber;
+	switch (moreOnNumberState) {
+		case 'three':
+			moreOnNumber = 3;
+			break;
+		case 'seven':
```
commuterjoy: I said to Max I think if you tested 3, 10 and 20 you'd see a more extreme reaction (i.e. quicker answer) between 3 focussed links and 10 quite focussed links and 20 not very focussed links. You could shuffle the order too so eliminate any bias.

PR #1235, opened 2016-02-11
Improve main image extraction

andygnewman:
Stop extracting images that are first child of elements with previous siblings.

Before;

screen shot 2016-02-11 at 09 07 23

After;

screen shot 2016-02-11 at 09 07 35


matthew-andrews:
:+1:

PR #1244, opened 2016-02-22
Add dummy version to package.json

quarterto:
We want to be able to `npm install Financial-Times/next-article` but npm doesn't like the lack of version. Currently we're working around it by installing with Bower which is just horrible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,6 @@
 {
   "name": "ft-next-article",
+  "version": "0.0.0",
   "private": true,
   "description": "Render a Next FT Article",
   "main": "app.js",


matthew-andrews:
are you not worried about the risk of doing this?  we could very easily break your code accidentally.

i'd really suggest copy+pasting for now:-
https://financialtimes.slack.com/archives/ft-next/p1456145904000125

quarterto:
We're [fixing](https://github.com/Financial-Times/google-amp/blob/master/bower.json#L6) to a particular known-good commit. Copy/paste is quite a painful option as we're using it for the XSLT stuff. We'd rather not take care of xsltproc ourselves.

quarterto:
(This pull request was unrelated to George's message; we're already pulling next-article in via Bower. Happy with copy/paste for the More On/From stuff)

PR #1245, opened 2016-02-22
Set valid entry point

keirog:


keirog:
Enables https://github.com/Financial-Times/next-build-tools/pull/340

PR #1250, opened 2016-02-24
Ads 705

leggsimon:
We now have `next-ads-api` which will get the dfpSite and dfpZone from ES and pass it directly to `next-ads-component`

@i-like-robots && @andygnewman


andygnewman:
Great to get some complexity out of the article app. Thanks.

leggsimon:
Build passes now if I can get a :+1:

andygnewman:
:+1:

PR #1252, opened 2016-02-26
This will be replaced with a more generic solution for everywhere

wheresrhys:
https://github.com/Financial-Times/n-layout/pull/48

May slightly break an old dashboard or two https://github.com/Financial-Times/next-beacon-dashboard/search?l=javascript&q=article&utf8=%E2%9C%93


wheresrhys:
@i-like-robots @andygnewman

PR #1256, opened 2016-03-02
rhr subhead and numbering mvt

andygnewman:
Styles the right hand rail recommended reads according to MVT variants ['sub-num', 'sub-none', 'none-num'].

sub-num
screen shot 2016-03-03 at 09 33 24

sub-none
screen shot 2016-03-03 at 09 33 47

none-num
screen shot 2016-03-03 at 09 34 18

control
screen shot 2016-03-03 at 09 34 36

and in case we have a result from the 'making links look like links' test
screen shot 2016-03-03 at 09 35 04

### DIFF
--- a/server/controllers/article.js
+++ b/server/controllers/article.js
@@ -138,10 +138,8 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		content.readNextTopic = content.primaryTag;
 
 		const rhrSubHeadNumber = res.locals.flags.articleRHRSubheadAndNumber;
-		if ( rhrSubHeadNumber && rhrSubHeadNumber !== 'control') {
-			content.rhrShowSubhead = rhrSubHeadNumber.split('-')[0] === 'sub' ? true : false;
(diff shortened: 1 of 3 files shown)


server/controllers/article.js
```diff
@@ -136,6 +136,12 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		);
 
 		content.readNextTopic = content.primaryTag;
+
+		const rhrSubHeadNumber = res.locals.flags.articleRHRSubheadAndNumber;
+		if ( rhrSubHeadNumber && rhrSubHeadNumber !== 'control') {
+			content.rhrShowSubhead = rhrSubHeadNumber.split('-')[0] === 'sub' ? true : false;
+			content.rhrShowNumber = rhrSubHeadNumber.split('-')[1] === 'num' ? true : false;
+		}
```
i-like-robots: Add a comment describing why this exists. Maybe use a regex.

i-like-robots:
```
/^sub-/
/-num$/
```

PR #1258, opened 2016-03-03
Remove colons from suggested reads text

mmockett:


mmockett:
@andygnewman Editorial don't like the colons in the Suggested Reads component so I've removed them.

matthew-andrews:
:scissors:

andygnewman:
:+1:

PR #1261, opened 2016-03-07
Don't let interactives break the Next design

matthew-andrews:
This stops the interactive breaking Next, but by breaking the interactive…

screen shot 2016-03-07 at 09 35 18

### DIFF
--- a/client/components/article/_main.scss
+++ b/client/components/article/_main.scss
@@ -159,6 +159,7 @@
 }
 
 .article__interactive {
+	max-width: 100%;
 	border: 0;
 }
 


commuterjoy:
+1 the assumption must be that things on the page are responsive

keirog:
We could be kinder to them and allow scrolling

matthew-andrews:
hmm, presumably they themselves could do that within the iframe? @keirog

keirog:
Probably not as easily as us removing `scrolling="no"` from the iframe itself. It could cause difficulties with mobile (scroll trap), but we do have a small outer gutter to ease that. Probably better than it being cut off, no?

matthew-andrews:
urgh … i don't understand these things

matthew-andrews:
@kavanagh @callumlocke would it be possible for you to fix the table itself?

matthew-andrews:
unfortunately i think `scrolling="no"` was there for a reason :/

kavanagh:
This wasn't meant to be a long term responsive solution and predates Next.

_Short answer:_

I can't think of a minor code change that will fix this. Your's is probably the most valid for now given the design of the article page.

_Long answer:_

These are some options to fix it properly. IG won't have time to look at any of them in the near future.

1. Don't make this an article. Put it in a page of it's own, without using an iframe for the table. It will work well enough without lots of effort.

2. Change the UI: don't use a table when there are a lot of mandatory columns. These don't work well on small screens. Think about a better solution for the Moocs page (and all the other ranking pages) and start from scratch.

3. next-article has a template that's suitable for wide content:
    - When the screen width allows it, let some elements be wider than the copy width (currently 670px). If the screen is less than this the elements are fluid to make best use of available space.
    - When the content has a minimum width and really does need to be wider than the screen you'll get a horizontal scrollbar on just the element that needs to be scrolled. This is adequate for the occasional spreadsheet type use case like this.

4. IG makes a template for wide content that does everything above, which would beg the obvious question... 
    - This option is dangerous because many people will want it but it's not sustainable for lots of articles.

5. Next could have an app that handles pages like this. There are a few of them - see rankings.ft.com

matthew-andrews:
Option 5 — we bounce people who visit these articles to the old FT.com page and leave sorting this out for a rainy day.

kavanagh:
:+1: didn't realise that's an option.

commuterjoy:
> we bounce people who visit these articles

Doesn't give anyone an incentive to fix it. 

Leaving it broken will mean people complain and it will get fixed.

matthew-andrews:
the only real fix is option 2 and that's not going to happen…  going back to FT.com will force it to be fixed just as urgently as leaving in the state it's currently in… that's going to go away eventually…

TheoLeanse:
Reopen if needed - This is a bit old and I think Rich Journalism will look at option 2.

PR #1262, opened 2016-03-07
Throw an error if hitting the article access controller with flag on

lc512k:
When the contentClassificationApi flag is on, access.js in next-content-classification-api should be used instead.

cc @matthew-andrews


matthew-andrews:
+1 good idea

PR #1263, opened 2016-03-10
un-silentify o-expander

TheoLeanse:
@i-like-robots @andygout 

to address this fellow:

image

Jira ticket is [here](https://jira.ft.com/browse/NFT-364)


andygout:
:+1:

PR #1270, opened 2016-03-16
remove interactive graphics that use http not https

andygnewman:
cc: @matthew-andrews 

Given our intention shortly to migrate the xslt transforms from article to ES, then this is probably where it should be done.


matthew-andrews:
+1

PR #1273, opened 2016-03-21
remove redundant flags params from xslt transform

andygnewman:
cc: @i-like-robots 

A bit of tidy up ahead of migrating xslt stylesheets to next-es-interface.

Flags are no longer required in the XSLT - only in cheerio transforms.


i-like-robots:
:ship:

PR #1275, opened 2016-03-21
Fix relative links to old 'theme' pages

i-like-robots:
Broken link reported on https://next.ft.com/content/7eaf4198-ec38-11e5-888e-2eadd5fbc4a4


server/transforms/relative-links.js
```diff
@@ -9,6 +9,7 @@ module.exports = function ($) {
 		el = cheerio(el);
 		if (el.attr('href')) {
 			el.attr('href', el.attr('href').replace(/https?:\/\/(?:www\.)?ft\.com\/topics\/places/, '/stream/regions'));
+			el.attr('href', el.attr('href').replace(/https?:\/\/(?:www\.)?ft\.com\/topics\/themes/, '/stream/topics'));
```
matthew-andrews: i think we could just delete all of these……

i-like-robots: Redirection doesn't seem to be working in this case (http://www.ft.com/topics/themes/Women_in_Business)

Issue #1276, opened 2016-03-23
Combine the figure-variant and external-images transforms


GlynnPhillips:
These transforms are reliant on each other and it makes sense to make them into one transform. Currently there are presentational variants calculated and used in both that are the same.

GlynnPhillips:
Done in #1282

PR #1277, opened 2016-03-23
Shrink wraps captions to the width of the image

GlynnPhillips:
Previously, tried to do this with css but it resulted in broken
responsive images.

We have ended up reverting to inline styles which is how it used to be
done when the transforms where XSLT and was lost during the move to JS.

See the [previous PR](https://github.com/Financial-Times/n-content-body/pull/7) for screenshots of how it will look.

I have just made changed to the external-image and figure trasnforms that means they are probably better off being combined now. Instead of bloating this change I have raised #1276 to deal with replacing and combining some of this code


i-like-robots:
looks good :ship:

PR #1280, opened 2016-03-23
Integrate Whitesource at deploy stage

matthew-andrews:
Integration of Whitesource into deploy step.

Runs against both node_modules (whitesource run) and bower_component (ws-bower) dependnencies.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ npm-debug.log
 .env
 .editorconfig
 .scss-lint.yml
+whitesource.config.json
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,17 @@ include n.Makefile
 
 TEST_APP := "ft-article-branch-${CIRCLE_BUILD_NUM}"
 
+install: whitesource.config.json
+	$(MAKE) install-super
+
+whitesource.config.json:
+	@if $(call IS_GIT_IGNORED); then echo '{ "apiKey": "$(WHITESOURCE_API_KEY)", "productName":"Next", "projectName":"$(call APP_NAME)" }' > $@ && $(DONE); fi
+
 coverage:
 	export apikey=12345; export api2key=67890; export AWS_SIGNED_FETCH_DISABLE_DNS_RESOLUTION=true; export NODE_ENV=test; istanbul cover node_modules/.bin/_mocha --report lcovonly 'test/server/**/*.test.js'
 
 unit-test:
-	$(NPM_BIN_ENV); export apikey=12345; export api2key=67890; export AWS_SIGNED_FETCH_DISABLE_DNS_RESOLUTION=true; export NODE_ENV=test; mocha 'test/server/**/*.test.js' --inline-diffs
+	export apikey=12345; export api2key=67890; export AWS_SIGNED_FETCH_DISABLE_DNS_RESOLUTION=true; export NODE_ENV=test; mocha 'test/server/**/*.test.js' --inline-diffs
 
 test:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -29,10 +35,15 @@ build-production:
 watch:
 	$(NPM_BIN_ENV); nbt build --dev --watch
 
-deploy:
+deploy: _deploy_whitesource
 	nbt deploy-hashed-assets
 	nbt ship -m
 
+_deploy_whitesource:
+	@$(NPM_BIN_ENV); whitesource run || echo "whitesource run failed, skipping"
```
matthew-andrews: if you update to the latest n-makefile (`make update-tools`) you won't need the `$(NPM_BIN_ENV)` stuff anymore

bower.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "next-article",
+  "name": "ft-next-article-bower",
```
matthew-andrews: **`-bower`**?

package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "ft-next-article",
+  "version": "0.0.0",
```
matthew-andrews: what's this for?

Makefile
andygnewman: Presume that applies to all @$(NPM_BIN_ENV) references in the Makefile?

bower.json
andygnewman: Whitesource overwrites it's store of node dependencies with bower dependencies if the name is the same in package.json and bower.json (or which ever is run last). Hence having a differential name in bower.json - presuming  the package.json name is the 'official' app name.
Does the bower.json name get used for anything?

package.json
andygnewman: Whitesource falls over if there's no version number https://github.com/whitesource/npm-plugin/blob/master/bin/ws_post.js#L113

bower.json
matthew-andrews: i don't think so this is fine, but it's a little bit brittle.  just read through their code it's a bit… hmm 
https://github.com/whitesource/bower-plugin/blob/master/bin/ws-bower.js

package.json
matthew-andrews: ah, would you mind raising a ticket against that repo to ask for that not to be an error — i think it's  reasonable for a consumer of npm modules to not have to have a version number.

in the mean time this is also ok +1

bower.json
andygnewman: Agree

matthew-andrews: the other thing about their code i'm not keen on is that it does a full bower install from scratch — which will give different results from what we've deployed.  i recon i could rework it to snapshot the **actual** dependencies that we shipped to production **and** make it faster

matthew-andrews: but that shouldn't blocked this PR ^^

package.json
andygnewman: https://github.com/whitesource/npm-plugin/issues/3

.gitignore
```diff
@@ -24,3 +24,7 @@ npm-debug.log
 .env
 .editorconfig
 .scss-lint.yml
+whitesource.config.json
+/.ws_bower
+ws*
+npm-shrinkwrap.json
```
matthew-andrews: might be able to avoid these by changing the whitesource commands to something like:-

```sh
(whitesource run || echo "whitesource run failed, skipping") && rm -r ws* && rm npm-shrinkwrap.json
```

PR #1282, opened 2016-03-30
Combines the image and figure transforms

GlynnPhillips:
As raised in #1276  this combines the external-image and
figure-variant transforms.

Not much has changed about the functionality, the code has just been
moved into one file. With moving the code into one transform its been
split into functions because it became a bit overwhelming.


andygnewman:
I remember now why I didn't combine these originally.

Not every `` is enclosed in a `<figure>`

Take the example of this [article] (https://next.ft.com/content/b8c42166-f640-11e5-803c-d27c7117d132)

The related article box half way down has an image without a figure, but still has the placeholding etc.

We could look at adding the figure element to the related box transform?

GlynnPhillips:
Thanks @andygnewman, I did wonder if this would be a scenario.

I imagine we could add the figure in this scenario or if it makes more sense we could just change this to initially loop over the images and look if it has a `figure` parent and if so make those changes.

I think I prefer the latter as its a better mirror of what we had before this change.

andygnewman:
I agree with you that the latter would seem the best approach.

GlynnPhillips:
@andygnewman this should now cover that scenario.

andygnewman:
:+1:  LGTM

PR #1283, opened 2016-03-31
Remove AB / MVT Test options

andygnewman:
- Make more-ons default to 3
- Remove link style from recommended reads


i-like-robots:
:ship:

PR #1286, opened 2016-04-01
open external links in new tab

andygnewman:
cc: @i-like-robots 
adds target=_blank to non-ft links in article body

### DIFF
--- a/server/transforms/external-links.js
+++ b/server/transforms/external-links.js
@@ -3,8 +3,8 @@
 module.exports = function($) {
 
 	const $links = $('a');
-	const httpMatcher = /^https?:\/\//;
-	const ftMatcher = /.ft.com\//;
+	const httpMatcher = /^http/;
+	const ftMatcher = /\.ft\.com\//;
 
 	$links.each(i => {
 		const $thisLink = $links.eq(i);
--- a/server/transforms/external-links.js
+++ b/server/transforms/external-links.js
@@ -1,0 +1,20 @@
+'use strict';
+
+module.exports = function($) {
+
+	const $links = $('a');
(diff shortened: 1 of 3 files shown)


server/transforms/external-links.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+module.exports = function($) {
+
+	const $links = $('a');
+	const httpMatcher = /^https?:\/\//;
+	const ftMatcher = /.ft.com\//;
```
i-like-robots: Need to escape `.`

server/transforms/external-links.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+module.exports = function($) {
+
+	const $links = $('a');
+	const httpMatcher = /^https?:\/\//;
```
i-like-robots: `^http` would suffice I think

test/server/transforms/external-links.test.js
```diff
@@ -0,0 +1,41 @@
...
+const expect = require('chai').expect;
+
+describe('External Links', () => {
+
+	it('opens external links in a new tab or window', () => {
+		let $ = cheerio.load(
+			'<a href="http://www.bloomberg.com/news/articles/2016-04-01/saudi-arabia-plans-2-trillion-megafund-to-dwarf-all-its-rivals">external link</a>'
+		);
+		const transformed$ = externalLinks($);
+		expect(transformed$('a').attr('target')).to.equal('_blank');
```
i-like-robots: the `transformed$` var isn't needed, the transform returns the original object so you can use `$('a')`

i-like-robots:
Looks good, just a couple of minor comments

PR #1287, opened 2016-04-01
swap out defunct fastft page from smoke tests

andygnewman:


matthew-andrews:
oops i broke this i guess

matthew-andrews:
+1

PR #1288, opened 2016-04-01
Bring your own ads

adgad:
@andygnewman @i-like-robots @VladDubrovskis @andrewgeorgiou1981 

Bring article ad slots and styles to within the app.


andygnewman:
You'll need to rebase this to master as we've updated the smoke test urls that article is based on.

server/transforms/body.js
```diff
@@ -44,6 +45,8 @@ module.exports = function (body, flags) {
 		.with(externalImagesEncoding)
 		.with(relatedBoxExpander)
 		.with(tableOfContents)
+		//TODO: write test here
+		.with(inlineAd)
```
andygnewman: We're soon (later this week) migrating some of these transforms to n-content-transform, so it can be reused by other apps who might want to get an article body, without the article app (which will be more related to the main presentation of the article page). 
I assume that inlineAd should remain in article when we do this? (and it means we can also use this to get an inline ad in Open In Stream.

views/article.html
```diff
@@ -2,9 +2,18 @@
...
 {{/defineBlock}}
+
 <article id="site-content" role="main" class="article{{#if designGenre}} article--brand {{#if designGenre.headshot}}article--brand--headshot{{/if}}{{/if}}" data-content-id="{{id}}" data-content-sources="{{#if articleV1}}v1{{/if}} {{#if articleV2}}v2{{/if}}" data-trackable="article-header">
-	<div class="article-header-ad-placeholder ad-placeholder"></div>
+{{#ifEquals adsLayout 'default'}}
+	{{> ads/banlb}}
+{{else}}
+	{{> ads/top}}
+{{/ifEquals}}
+
```
andygnewman: I notice you haven't included it into podcasts view?
However help is at hand as I'm getting this [PR] (https://github.com/Financial-Times/next-article/pull/1185) in soon, which uses a common page structure for podcast and article content. So you might want to wait and rebase onto that.

server/transforms/body.js
adgad: Interesting. Haven't considered open-in-stream yet, but sounds like a good thing.

Is n-content-body for use in e.g. amp as well? (where they need a custom element for ads)?

views/article.html
adgad: Indeed an oversight. I can wait and rebase :)

adgad:
@andygnewman Updated this with all the new changes

adgad:
The ad slot appears in the fragment view - but it won't be initialised yet on stream pages (since it's brought in after the page has loaded). Can ignore this for now until we look at stream pages in more detail.

views/partials/right-hand/right-hand.html
```diff
@@ -1,6 +1,13 @@
 <div class="article__aside o-grid-remove-gutters--XL" data-o-grid-colspan="0 L4 XL3 XLoffset1">
 	<div class="sidebar-ad-placeholder ad-placeholder"></div>
```
andygnewman: Do you still need this in here given lines 5-9 in the new version?

adgad: Nope, will delete

views/partials/right-hand/right-hand.html
```diff
@@ -1,6 +1,13 @@
...
 	{{{tocHtml}}}
+
```
andygnewman: Ads were previously above any toc (Table of Contents) - which don't appear often.

An example of a page with ToC is https://next.ft.com/content/dbbe9112-940a-11e5-b190-291e94b77c8f

Just wondered whether this was a conscious decision to have this happen.

adgad: Ah thanks, I wasn't sure were ToCs were in the code before the refactor! The ad should be above.

andygnewman: I think they were below the Recommended Reads, which didn't make sense, so I moved them above.

andygnewman:
👍

PR #1289, opened 2016-04-05
Updates heading levels and lists on related content

GlynnPhillips:
Changing the "Read latest" heading from a `<p>` to a `<h3>`. This makes it consistent
between header and bottom box and it should be a heading.

Removes the `<h4>` from  within the "read latest" box as they should be
links not headings.

Removes the `<h4>` from the  "More of this topic" box as it should be a
list of links not headings.

Put all items in "More on this topic" box into a list.

Changes More on this topic subheadings from `<div>` to `<p>`

### DIFF
--- a/client/components/read-next/main.scss
+++ b/client/components/read-next/main.scss
@@ -72,6 +72,8 @@
 	@include oTypographySans(xs);
 	text-transform: uppercase;
 	color: oColorsGetColorFor(timestamp, text);
+	display: block;
+	margin-top: 5px;
 }
 
 .next-up__timestamp__more-recent {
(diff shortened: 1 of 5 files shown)


andygnewman:
Looks sensible.

Have you checked that all the variants for our current MVT on the suggested reads (https://toggler.ft.com/ - articleRHRSubheadAndNumber) are unaffected?

GlynnPhillips:
Yeah I think I understood them correctly and here is each variant now

screen shot 2016-04-05 at 09 24 24
screen shot 2016-04-05 at 09 23 32
screen shot 2016-04-05 at 09 22 52
screen shot 2016-04-05 at 10 29 15

andygnewman:
👍

PR #1290, opened 2016-04-05
encode srcset aswell as src for images

andygnewman:
cc: @i-like-robots


i-like-robots:
I have literally no clue but I trust it works! :ship:

PR #1299, opened 2016-04-07
Fix targeting name for inline ad

adgad:
/cc @VladDubrovskis @andygnewman 

* Update the pos names, without affecting live site pos names.
* Remove URL mapping hack
* clear both (probably needs more thought) on demo responsive ads layout, to prevent ad appearing squashed next to an aside.

### DIFF
--- a/server/transforms/inline-ad.js
+++ b/server/transforms/inline-ad.js
@@ -1,8 +1,7 @@
 module.exports = function ($, flags, adsLayout) {
 	const pars = $('p');
-	console.log('adsLayout', adsLayout);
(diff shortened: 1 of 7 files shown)


server/transforms/inline-ad.js
```diff
@@ -1,12 +1,13 @@
-module.exports = function ($) {
+module.exports = function ($, flags, adsLayout) {
 	const pars = $('p');
+	console.log('adsLayout', adsLayout);
```
VladDubrovskis: console.log?

VladDubrovskis:
:+1:

PR #1300, opened 2016-04-08
Move the error for content classfication logic when content classifca…

matthew-andrews:
…tion api flag is swtiched on to only run for access requests


matthew-andrews:
note to self: squash this

PR #1303, opened 2016-04-11
Handle no more ons

ironsidevsquincy:
Causes e.g. - https://next.ft.com/content/6d62306e-fc08-11e5-b3f6-11d5706b613b

@i-like-robots, @andygnewman


andygnewman:
👍 , though worrying that it has no metadata, I'm sure a quick reingest would sort that out, but still good to catch this.

PR #1304, opened 2016-04-11
Swap over to use bodyHTML from ES

andygnewman:
To Do
- [x] Remove XSLT transforms, set up and tests
- [x] Remove redundant Cheerio transforms and tests
- [x] Apply n-content-transform to bodyHTML
- [x] Default to bodyXML if no bodyHTML
- [x] Apply remaining 'extremely opinionated' article transforms (refactored these to account for flags and other settings penitent to articles)


server/controllers/article.js
```diff
@@ -67,13 +63,20 @@ module.exports = function articleV3Controller(req, res, next, content) {
...
-			content.bodyHtml = fragments.bodyHtml;
-			content.tocHtml = fragments.tocHtml;
-			content.mainImageHtml = fragments.mainImageHtml;
-		})
-	);
+	// Apply content and article specific transforms to bodyHTML
+	if (content.bodyHTML) {
+		let transformedBody = transformArticleBody(content.bodyHTML, res.locals.flags);
+		content.bodyHTML = transformedBody.bodyHTML;
+		content.tocHTML = transformedBody.tocHTML;
```
i-like-robots: this is interface is actually a bit weird... put one string in get object with three properties back. 

Maybe that can be more cleanly integrated: 

```js
Object.assign(content, transformArticleBody(content.bodyHTML, res.locals.flags))`
```

server/controllers/article.js
```diff
@@ -25,15 +27,9 @@ function isCapiV2(article) {
...
-		id: article.id,
-		webUrl: article.webUrl
-	};
-
-	return articleXsltTransform(article.bodyXML, 'main', xsltParams).then(articleBody => {
-		return bodyTransform(articleBody, flags, article.adsLayout);
-	});
+function transformArticleBody(bodyHTML, flags) {
+	const articleBody = contentTransform(bodyHTML, flags);
+	return articleTransform(articleBody, flags);
```
i-like-robots: I think the naming of `transformArticleBody`, `contentTransform` and `articleTransform` need to be a little more distinctive, it's hard to process

server/transforms/extract-main-image-and-toc.js
```diff
@@ -2,27 +2,27 @@
 
 module.exports = function ($) {
```
i-like-robots: should this be separate modules? One responsibility and that

server/controllers/article.js
andygnewman: I would suggest;
`transformArticleBody` (or `transformBody`) remains the name of the function in article.

the `n-content-transform` could be `nContentTransform` or `opinionatedContentTransform`

the final article specific transform could be 'articleAppContentTransform` or `appSpecificContentTransform`

Any preferences?

server/controllers/article.js
andygnewman: Good idea with Object.assign

server/transforms/extract-main-image-and-toc.js
andygnewman: Good point - especially as I can foresee a time when we just get mainImage from the mainImage attribute rather than stripping out of body.

server/controllers/article.js
```diff
@@ -55,13 +50,23 @@ module.exports = function articleV3Controller(req, res, next, content) {
...
-			content.bodyHtml = fragments.bodyHtml;
-			content.tocHtml = fragments.tocHtml;
-			content.mainImageHtml = fragments.mainImageHtml;
-		})
-	);
+	// If no bodyHTML, revert to using bodyXML
+	content.bodyHTML = content.bodyHTML || content.bodyXML;
+
+	// Apply content and article specific transforms to bodyHTML
+	if (content.bodyHTML) {
```
i-like-robots: This will always be true because of the line above.

server/controllers/article.js
```diff
@@ -55,13 +50,23 @@ module.exports = function articleV3Controller(req, res, next, content) {
...
+				flags: res.locals.flags,
+				fragment: req.query.fragment,
+				adsLayout: content.adsLayout,
+				usingBodyXML: !content.bodyHTML && content.bodyXML
+			}
+		));
+	}
+
+	// Add the inline ad placement to bodyHTML once determined whether to use HTML or XML
+	content.bodyHTML = inlineAdTransform(content.bodyHTML, content.adsLayout);
```
i-like-robots: Should this simply be included within the above... or is the assumption we only want to apply this to both `bodyHTML` _and_ fallback `bodyXML`?

server/controllers/article.js
```diff
@@ -55,13 +50,23 @@ module.exports = function articleV3Controller(req, res, next, content) {
...
+	// If no bodyHTML, revert to using bodyXML
+	content.bodyHTML = content.bodyHTML || content.bodyXML;
+
+	// Apply content and article specific transforms to bodyHTML
+	if (content.bodyHTML) {
+		Object.assign(content, transformArticleBody(content.bodyHTML, {
+				flags: res.locals.flags,
+				fragment: req.query.fragment,
+				adsLayout: content.adsLayout,
+				usingBodyXML: !content.bodyHTML && content.bodyXML
```
i-like-robots: This will always be true too, what does the flag do?

server/transforms/extract-main-image.js
```diff
@@ -1,28 +1,26 @@
...
 
 	// find image or slideshow in the body
 	const $firstMainImage = $('figure.n-content-image--full, figure.n-content-image--center, ft-slideshow').eq(0);
-	let mainImageHtml;
+	let mainImageHTML;
 
 	// check that it is the first element in the body
 	if (
 		$firstMainImage.length &&
 		!$firstMainImage.prev().length &&
```
i-like-robots: If we write this as `$firstMainImage.prev().length === 0` it will read a little nicer =]

server/transforms/extract-toc.js
```diff
@@ -0,0 +1,13 @@
...
+const cheerio = require('cheerio');
+
+module.exports = function (bodyHTML) {
+
+	const $ = cheerio.load(bodyHTML);
+
+	return {
+		tocHTML: $.html($('.table-of-contents').remove()),
+		bodyHTML: $.html()
+	}
```
i-like-robots: bad linter

server/transforms/inline-ad.js
```diff
@@ -1,12 +1,15 @@
-module.exports = function ($, flags, adsLayout) {
+const cheerio = require('cheerio');
+
+module.exports = function (body, options) {
+	const $ = cheerio.load(body, { decodeEntities: false });
 	const pars = $('p');
 	pars.each((index, par) => {
 		if(index > 1 && par.next && par.next.name === 'p') {
```
i-like-robots: I think this could also be expressed as: 

```js
$('p')
// Ignore the first
.slice(1)
// Select all next sibling paragraphs
.next('p')
// Choose the first
.first()
// Insert content before
.before()
```

server/controllers/article.js
i-like-robots: Maybe we should have `bodyTransformEnhanced` and `bodyTransformAll`

server/controllers/article.js
andygnewman: Yes, ended up moving this logic around and forgot to clean up.

server/controllers/article.js
andygnewman: As your comment below, this is so any transforms can decide whether to operate or not depending on whether it is the prime HTML or we've fallen back to V1 only body. Eg. don't run related content expander or table of contents on V1, but do run ads and light sign-up.

server/controllers/article.js
andygnewman: Yes, it was included above, but not cleared out of here. I had gone with a model of the individual transform controlling whether it should operate on the fallback XML body or not, rather than have distinct aggregated transforms for all & enhanced. Get the point and can change it to that. I'll pop over later and discuss.

server/transforms/inline-ad.js
andygnewman: Not quite sure that would work, as there's still the case of it having to be between two p's, but also there's a new condition to avoid putting it in asides.

Like the idea of using slice to avoid having to do the index > 1 condition though.

Will mention this to Arjun

i-like-robots: > Not quite sure that would work, as there's still the case of it having to be between two p's

that's how what the next sibling is doing =]

i-like-robots: https://tonicdev.com/56d5a9a467c99d0d00e23758/5718b1eea1e962110076c03c

andygnewman:
Agreed that we should run all article specific transforms regardless of origin of body and just pass the standard flags through. Need to make these changes before merging.

PR #1306, opened 2016-04-12
Force the curl's against FT.com to stay on FT.com.

adgad:
cc @adambraimbridge @wheresrhys


wheresrhys:
oooh nasty. If you're on bug duty, this will probably need adding to falcon redirect and maybe next redirector too

PR #1307, opened 2016-04-12
Add a light signup component after an article's 5th paragraph

TheoLeanse:
@andygnewman @i-like-robots


server/transforms/light-sign-up.js
```diff
@@ -0,0 +1,36 @@
...
+							<div class="n-light-signup__close"></div>
+							<h2 class="n-light-signup__heading">Get a free taste of the FT</h2>
+							<p class="n-light-signup__tagline">
+								Sample the FT, with the top stories sent FREE to your inbox every morning. For 7 days, you’ll get an email with free access to the 3 most read FT stories.
+							</p>
+						</div>
+					</div>
+					<div class="n-light-signup__secondary">
+						<div class="o-grid-row o-grid-row--compact">
+							<div data-o-grid-colspan="12">
```
i-like-robots: Is there any use in these grid column elements?

test/server/transforms/light-sign-up.test.js
```diff
@@ -0,0 +1,58 @@
+/*global describe, it*/
+"use strict";
+var cheerio = require('cheerio');
+var expect = require('chai').expect;
+var lightSignupTransform = require('../../../server/transforms/light-sign-up');
+
+describe('Light Signup component inside body', function () {
+
+	const lightSignupHtml = `<div class="n-light-signup" data-trackable="light-signup">
```
i-like-robots: Could this fixture be simplified to reduce noise? E.G. simply `<div class="n-light-signup">$$$</div>`

server/transforms/light-sign-up.js
```diff
@@ -0,0 +1,36 @@
+module.exports = function ($, flags) {
+	if (!flags.lightSignUp) return $;
+
+	const pars = $('p');
```
i-like-robots: Maybe switch to `p:nth-of-type(3)` so we don't have to loop

test/server/transforms/light-sign-up.test.js
```diff
@@ -0,0 +1,58 @@
...
+									<p class="n-light-signup__no-spam">Unsubscribe at any time. <a href="#" class="n-light-signup__link" data-trackable="see-sample-email">See sample email</a>.</p>
+								</form>
+							</div>
+						</div>
+					</div>
+				</div>`;
+
+	it('should not do anything if the lightSignUp flag is off', function () {
+		var $ = cheerio.load('<p>1</p><p>2</p><p>3</p><p>4</p><p>5</p><p>6</p>');
+		$ = lightSignupTransform($, {});
```
i-like-robots: no need to reassign, cheerio acts upon a reference and doesn't return a new object

server/transforms/light-sign-up.js
andygnewman: The logic here is to search for a gap between two p's (rather than between a p and an image for example) which could happen p:nth-of-type were used.

i-like-robots: Ahh good point... fortunately that can still be expressed:

```js
$('p:nth-of-type(3) ~ p:has(+p)')

// or

$('p:nth-of-type(3) ~ p').filter(':has(+p)')
```

andygnewman: Nice

i-like-robots: Bit weird since the CSS selectors get converted to xpath and I don't think we can provide the xpath expression directly.

server/transforms/light-sign-up.js
TheoLeanse: nope - removed. thanks!

TheoLeanse:
@andygnewman @i-like-robots 
better?

server/transforms/light-sign-up.js
i-like-robots: I've tested this out here https://tonicdev.com/56facff6812b1112005f27c1/570e955d1f92ef1300eead7b

i-like-robots: Without the parent wrapper this can't be expressed with a CSS selector alone but it can still be done:  https://tonicdev.com/56facff6812b1112005f27c1/570ea0a0d7a6ec1100517023

```js
cheerio = require("cheerio")
html = '<p>hello</p><p>world</p><p>foo</p><p>bar</p><div>oh no</div><p>baz</p><p>qux</p>'
$ = cheerio.load(html)

$('p').slice(3).next('p').first().html() // qux
```

TheoLeanse: nice.

TheoLeanse: with that second option (without parent wrapper), how would you prevent the component from being added before an element that isn't a `<p>`?

i-like-robots:
Much better having the template externally, far less HTML being repeated 😃 

Squerge!

server/transforms/light-sign-up.js
i-like-robots: Oh, with the example above you'd have to insert your placholder _before_ the selected node, not after =)

TheoLeanse: aha!

On 14 April 2016 at 10:56, Matt Hinchliffe <notifications@github.com> wrote:

> In server/transforms/light-sign-up.js
> <https://github.com/Financial-Times/next-article/pull/1307#discussion_r59690724>
> :
>
> > @@ -0,0 +1,36 @@
> > +module.exports = function ($, flags) {
> > +	if (!flags.lightSignUp) return $;
> > +
> > +	const pars = $('p');
>
> Oh, with the example above you'd have to insert your placholder *before*
> the selected node, not after =)
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/Financial-Times/next-article/pull/1307/files/b656a143e0e679aa241e3f9bec2718e19d1b6b36#r59690724>
>

-- 

------------------------------

*This email was sent by a company owned by Financial Times Group Limited 
("FT Group <http://aboutus.ft.com/corporate-information/#axzz3rajCSIAt>"), 
registered office at Number One Southwark Bridge, London SE1 9HL.  
Registered in England and Wales with company number 879531. This e-mail may 
contain confidential information. If you are not the intended recipient, 
please notify the sender immediately, delete all copies and do not 
distribute it further.  It could also contain personal views which are not 
necessarily those of the FT Group.  We may monitor outgoing or 
incoming emails as permitted by law.*

i-like-robots:
Eeek! Please squash in future (sorry, squerge = squash and merge) to keep a clean history on master =]

TheoLeanse:
Oh I'm sorry, didn't understand! 
I can't retroactively squerge?

i-like-robots:
Not to worry, my bad for using dumb jargon ([see here](http://remy.bach.me.uk/blog/2015/02/the-humble-squerge/))!

It's possible to re-write Git history (`git rebase -i`) but that should be restricted to branches you're working on alone so other people don't get stuck in limbo.

PR #1309, opened 2016-04-14
Don't wait for load event before initialising comments

ironsidevsquincy:
At this point there's no guarantee the load event hasn't already fired (seems to be more common on Safari), hence comments will never load

Any idea why this was here in the first place? Seems no reason the 'wait' for the load event - @i-like-robots @andygnewman


i-like-robots:
No clue... I suppose if there is/was a reason we could check `document.readyState`

andygnewman:
No idea why there was an event listener for initiating this as it is as dependent on the DOM loading as the rest of the inits that precede it which didn't have it.

PR #1313, opened 2016-04-18
change copyright from a p to a footer element

andygnewman:
To aid standardised styling of footer elements in article / o-typography


views/partials/copyright-notice.html
```diff
@@ -1,2 +1,2 @@
-<p class="n-content-copyright"><a href="http://www.ft.com/servicestools/help/copyright" data-trackable="link-copyright">Copyright</a>
-The Financial Times Limited {{thisYear}}. All rights reserved. You may share using our article tools. Please don't cut articles from FT.com and redistribute by email or post to the web.</p>
+<footer class="n-content-copyright"><a href="http://www.ft.com/servicestools/help/copyright" data-trackable="link-copyright">Copyright</a>
```
i-like-robots: wrap in a `<small>` too - it was redefined as having small print semantics in HTML5.

PR #1314, opened 2016-04-18
Lazy loads comments js and css

GlynnPhillips:
Prevents users who never scroll down to the comments from downloading
the extra js and css.

For now just putting this behind a flag so Todd can easily access it to work on the design

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,12 +27,16 @@ run:
 	nbt run
 
 build:
+	nbt build --main-js comments.js --main-sass comments.scss --skip-about --skip-haikro --skip-hash --dev
 	nbt build --dev
 
 build-production:
+	nbt build --main-js comments.js --main-sass comments.scss --skip-about --skip-haikro --skip-hash
 	nbt build
 
+
 watch:
+	@echo " **************************************************\n WATCH IS PARTIALLY BOKEN\n Within next-article there are multiple bundles\n NBT watch will only rebuild main.js and main.scss\n **************************************************"
 	nbt build --dev --watch
 
 deploy: _deploy_whitesource
--- a/client/components/comments/skeleton.js
+++ b/client/components/comments/skeleton.js
@@ -1,0 +1,30 @@
+'use strict';
+
+const createSkeleton = {
+	init: function() {
+		const skeleton = document.createElement('div');
+		const skeletonInner = document.createElement('div');
+		const skeletonPod = document.createElement('div');
+		const skeletonEdit = document.createElement('div');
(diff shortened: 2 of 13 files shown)


Makefile
```diff
@@ -27,12 +27,16 @@ run:
...
+	nbt build --main-js comments.js --main-sass comments.scss --skip-about --skip-haikro --skip-hash --dev
 	nbt build --dev
 
 build-production:
+	nbt build --main-js comments.js --main-sass comments.scss --skip-about --skip-haikro --skip-hash
 	nbt build
 
+
 watch:
+	@echo " **************************************************\n WATCH IS PARTIALLY BOKEN\n Within next-article there are multiple bundles\n NBT watch will only rebuild main.js and main.scss\n **************************************************"
```
adambraimbridge: Yeah, watch is also meant to rebuild when partials are edited, but it doesn't.

client/components/comments/skeleton.js
```diff
@@ -0,0 +1,30 @@
...
+		skeleton.appendChild(skeletonInner);
+		skeletonInner.appendChild(skeletonPod);
+		skeletonInner.appendChild(skeletonEdit);
+		skeletonInner.appendChild(skeletonFollow);
+		skeletonInner.appendChild(skeletonSubmit);
+		commentsEl.parentNode.insertBefore(skeleton, commentsEl.nextSibling);
+	}
+}
+
+module.exports = createSkeleton;
```
adambraimbridge: I've never used <template> tags, but they _might_ be useful here? https://developer.mozilla.org/en/docs/Web/HTML/Element/template ... maybe not.

laurieboyes: IE :rage1: 

screen shot 2016-04-18 at 14 10 00

Makefile
ironsidevsquincy: `watch` is just for front-end assets. `nodemon` should restart when a partial is changed, but doesn't. not sure why

out of scope, but switching to `n.makefile` would fix this watch

GlynnPhillips: This is using `n.makefile`, how could this be changed to fix this @ironsidevsquincy?

I think @i-like-robots had a plan for this potentially being replaced in the long term when articles are picked back up

ironsidevsquincy: sorry, more specifically, n.makefile's `watch`, which uses webpack's watch directly (which has a far more intelligent watching system)

i-like-robots: Yep, we need to update article to use Webpack instead of OBT. It's not that hard... if I get 15 mins today I will do it!

PR #1319, opened 2016-04-18
Use Arial for italic Metric

ironsidevsquincy:


andygnewman:
Should change the n-content... styles in n-content-body

[images](https://github.com/Financial-Times/n-content-body/blob/master/scss/_images.scss)

[tags](https://github.com/Financial-Times/n-content-body/blob/master/scss/_tags.scss#L58)

I'm trying to reconcile a lot of n-content-body with o-typography so if it's there there's more chance of me migrating it.

client/main.scss
```diff
@@ -73,3 +73,11 @@ $o-comments-is-silent: false;
...
 	}
 }
+
+// stop italic metric downloading
+.n-content-image__caption,
+.n-content-tags__product-tour-link i,
+.article__share i {
+	font-family: Arial, sans-serif;
+	font-size: 0.8em;
+}
```
i-like-robots: Should we put this in a `_hacks.scss` or `_shame.scss` partial so it's flagged for refactoring?

ironsidevsquincy:
Actually, now using faux italic for our italic needs (https://github.com/Financial-Times/next-sass-setup/pull/131), so closing

PR #1323, opened 2016-04-19
Fix for ad appearing inside an aside box

adgad:
/cc @andygnewman 

Amends the logic for the in-article ad to only look at top-level p tags (which I _think_ is a reasonable assumption).

Before:
screen shot 2016-04-19 at 14 19 57

After:
screen shot 2016-04-19 at 14 19 46


andygnewman:
Logic looks good to me. 👍

PR #1325, opened 2016-04-19
Cache for an hour

wheresrhys:
Any reason not to?


matthew-andrews:
i think we need to implement purging…… otherwise editorial gonna be a bit angry i think

matthew-andrews:
really think we could start off figuring out purge+long cache logic problems in content-classification-api first :]

andygnewman:
What would happen if Editorial update an existing story - does that override the cache?

wheresrhys:
How about we cache for longer if a story is more than 24/48/72 hours long?

matthew-andrews:
@andygnewman no not without us implementing purging, which i don't trust on next.ft.com atm

PR #1326, opened 2016-04-19
only cache related link endpoints for longer

wheresrhys:
@andygnewman @matthew-andrews Second attempt. Surely editorial can't object to related links being cached for an hour?

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -31,7 +31,7 @@ app.post('^/preview$', bodyParser.json(), require('./controllers/preview'));
 
 // Apply this after the preview controller. Previews should not be cached
 app.use((req, res, next) => {
-	res.set('Surrogate-Control', 'max-age=600,stale-while-revalidate=20,stale-if-error=259200');
+	res.set('Surrogate-Control', 'max-age=3600,stale-while-revalidate=20,stale-if-error=259200');
 	next();
 });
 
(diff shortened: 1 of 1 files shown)


matthew-andrews:
seems ok to me!

andygnewman:
I'm going to be contrarian here. I'm presuming that the objective is to find the right balance between performance / ability to serve from cache and ensuring the reader has the most up to date news.

In which case I'd like to argue for a different approach.

At the moment we use the content uuid in the url for all the related content pieces. Which presumably means a different cache for each instance of the uuid.

However the uuid is rarely used in generating the related content. It's something that's been in the back of my mind to refactor out. Thus if we didn't use the uuid, we would likely increase the chances of being able to serve from a cache.

Eg. story packages are often shared across articles, more-on topics may be similar, and special reports don't need the uuid.

It would be good to try this approach and see what improvement that brings us - do we have reporting on % of requests served by cache? Doubt it would bring us the same level of benefit as taking the cache up to 60 mins, but might halve it, meaning we need not be so severe in our cache reduction.

Thoughts?

wheresrhys:
you're definitely on to something there. So the idea would be instead of requesting `more-ons/{story uuid}` we request `more-ons/{primaryTheme id}` etc... is that what you're getting at? I'd say that would have a much greater impact than what I'm doing in this PR.

In the meantime, are you happy for me to merge this?

andygnewman:
Yes, that's the idea, but just been having another look at the code and we're using the parent uuid to filter out the parent story from both the more-ons and the special reports. Can still remove it from the story package though. I'll raise an issue for that one.

We could move the deduping out of the specific controller and put it client side so we can take advantage of caching, but that would increase the client side file size / processing, so not sure if that is a good thing or not.

In the mean time OK for you to merge this.

PR #1327, opened 2016-04-19
Light signup mvt test

TheoLeanse:
@i-like-robots @andygnewman @commuterjoy 

We're going to test the placement of this light signup component, so changing the transform accordingly.

(Matt H, I promise to squerge if OK)


server/transforms/light-sign-up.js
```diff
@@ -1,13 +1,34 @@
...
-	pars.each((index, par) => {
-		if(index > 3 && par.next && par.next.name === 'p') {
-			$(par).after('<div class="n-light-signup__container"></div>');
-			return false;
-		}
-	});
+
+	if (variant === 'control') return $;
+
+	if (variant === 'top') {
```
i-like-robots: Is it worthing DRY-ing this up, maybe by assigning the position to a variable?

```js
let position;

switch (flags.lightSignUp) {
  case: 'top':
    position = 0;
  case 'mid':
    position = 3;
  case 'end':
    position = pars.length;
}
```

TheoLeanse:
@i-like-robots I've tried to get it a bit drier, but perform the check for a next paragraph only on the top and mid variants.

i-like-robots:
🚢

PR #1334, opened 2016-04-21
Email article

bjfletcher:
This allows users to click on a button on a paywalled article and share the article with others through email - as a gift if necessary.

This uses the new [n-email-article](https://github.com/Financial-Times/n-email-article) component.

This branch is work in progress. I will write again in here when this is ready for review.


client/components/share/main.js
```diff
@@ -28,4 +30,9 @@ exports.init = function() {
 		new OShare(shareContainer);
 		loadShareCount();
 	}
+
+	const button = document.querySelector('[data-n-article-email-clickable]');
```
andygnewman: I wonder whether we should be more specific with this name. We might in future have other buttons that we want to handle in a similar way.

package.json
```diff
@@ -21,6 +21,8 @@
     "n-health": "^1.3.0",
     "n-podcast-mapping": "Financial-Times/n-podcast-mapping",
     "next-ft-api-client": "^3.16.0",
+    "react": "^15.0.1",
```
andygnewman: Does react need to be a dependency of article or can it just be a dependency of n-email-article?

PR #1338, opened 2016-04-25
Non-body content html structure

andygnewman:
Improve consistency of heading tags in article - see this [issue](https://github.com/Financial-Times/next-es-interface/issues/383

Covering:
- more-ons
- story package
- special reports
- read next (top & bottom, rhr already fixed)


views/partials/header/read-next-header.html
```diff
@@ -2,10 +2,14 @@
 	<aside class="article__aside o-grid-remove-gutters--XL" data-o-grid-colspan="hide L4 XL3 XLoffset1" data-trackable="next-up-header" role="complementary" aria-hidden="true">
 		<div class="next-up next-up__header">
 			<h3 class="next-up__intro next-up__intro__header">{{#if moreRecent}}Read latest:{{else}}Read next:{{/if}}</h3>
-			<a href="{{url}}" class="next-up__headline" data-trackable="headline-link--{{source}}{{#if moreRecent}}--moreRecent{{/if}}">{{title}}</a>
-			<time class="next-up__timestamp {{#if moreRecent}}next-up__timestamp__more-recent{{/if}} o-date" data-o-component="o-date" datetime="{{#dateformat}}{{publishedDate}}{{/dateformat}}" data-o-date-js>
-				{{#dateformat "dddd, d mmmm, yyyy"}}{{publishedDate}}{{/dateformat}}
-			</time>
+			<article>
+				<h4>
```
i-like-robots: I think this was OK previously, the article element has semantics on paper but not in practise and the h3 does a decent job of describing this segment.

if we wanted this to really mean something we'd have to drop in some schema.org

views/related/special-report.html
```diff
@@ -8,16 +8,16 @@
...
 		</div>
-		<ol class="stream-card__articles">
+		<div class="stream-card__articles">
 			{{#each articles}}
-				<li class="stream-card__article" data-trackable="article">
+				<article class="stream-card__article" data-trackable="article">
 					<time data-o-component="o-date" class="stream-card__article__date o-date" datetime="{{#dateformat}}{{publishedDate}}{{/dateformat}}">
 						{{#dateformat "dddd, d mmmm, yyyy"}}{{publishedDate}}{{/dateformat}}
 					</time>
 					<br />
```
i-like-robots: unnecessary line break

views/partials/header/read-next-header.html
andygnewman: Was trying to reconcile it with [Suggested Reads](https://github.com/Financial-Times/next-article/blob/master/views/partials/right-hand/suggested-reads.html#L16) implementation.

Would we therefore use it when in a list format (eg. suggested reads, story package, more-ons), but not on a single instance implementation (eg. read next, related article promo-box)?

i-like-robots: I think the distinction is that a heading should describe some content - so the suggested reads preceeds a paragraph of text

i-like-robots:
👍  squerge

PR #1339, opened 2016-04-25
Use `srcset` and `sizes` rather than <picture />

ironsidevsquincy:
Also tidy the main image sizes

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -3,7 +3,6 @@
   "dependencies": {
     "labs-o-share": "ftlabs/labs-o-share#^v3.1.0",
     "n-content-body": "^2.0.0",
-    "n-image": "^2.0.0",
     "n-layout": "^6.0.0",
     "n-message-prompts": "^2.0.0",
     "n-video": "^3.2.0",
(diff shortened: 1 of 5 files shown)


andygnewman:
Looks OK, but wonder why you've changed the image sizes of 320 and 480 to 300 and 500? Whilst they may look more convenient, the 320 and 480 related to explicit use cases on the site.

ironsidevsquincy:
As far as I can see they're not used - `M`: 700, `L`: 620, `XL`: 680. Anything smaller is full width of the viewport, so these sizes seemed to cover a wide range

andygnewman:
Sorry, should have said that they related to more-ons, though usage may have changed.

ironsidevsquincy:
Might have got it wrong, but thought the code in `article-pod-mapping-v3.js` related to the more ons. Just removed the 470px width image from that, as no images are shown on smaller breakpoints

andygnewman:
Yeah, sorry was getting my srcsets and sizes confused. 👍

ironsidevsquincy:
ha, yeah, the terminology is all a bit ambiguous/confusing

PR #1340, opened 2016-04-27
unvarying where appropriate

wheresrhys:
@matthew-andrews @andygnewman

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "test": "make test -j 3"
   },
   "dependencies": {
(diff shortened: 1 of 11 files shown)


server/controllers/article.js
```diff
@@ -115,6 +115,7 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		.then(() => {
 			content.contentType = 'article';
 			if (req.query.fragment) {
+				res.unvaryAll('wrapper');
```
wheresrhys: not sure about this one

PR #1342, opened 2016-04-27
pinning cheerio to 19 - cause of memory leak?

wheresrhys:


andygnewman:
Happy to see this tested, but should it be a tilde ~ to pin it to 0.19.0?
We're using it in [n-content-transform](https://github.com/Financial-Times/n-content-transform/blob/master/package.json#L6) as well. Should we pin this back or wait for the outcome of testing this? We're not consuming n-content-transform in article yet, but have [PR](https://github.com/Financial-Times/next-article/pull/1304) ready to go to do so, but can hold off until we better understand this memory leak.

wheresrhys:
below 0, ^ behaves the same as ~ in npm. I have a (possibly optimistic) belief that using ^ means people are less likely to correct the 'mistake'. I'll let you know if this pr fixes the memory issue

PR #1343, opened 2016-04-28
Fix empty space next to aside

adgad:
@VladDubrovskis @andygnewman @andrewgeorgiou1981 @leggsimon 

* Stop empty ad taking up space in between article
* If an ad was next to an aside, wait another paragraph, so there is less empty space next to the aside

After:
screen shot 2016-04-28 at 09 29 38


VladDubrovskis:
👍

PR #1345, opened 2016-04-28
Fix mobile ads

adgad:
/cc @leggsimon @andrewgeorgiou1981 @VladDubrovskis


VladDubrovskis:
👍

adgad:
Checked this on IE9 and it looks okay (plus its one line of CSS) so gonna merge

PR #1357, opened 2016-05-05
Craig shuffle

andygnewman:
Another ad-blocking response variant

Shuffles the characters inside a word.

screen shot 2016-05-05 at 12 21 45


andygnewman:
Actually I'm going to do this differently - add the two files onto a new branch off master to get rid of all the terrible history.

PR #1366, opened 2016-05-06
Remove serif bold usage

ironsidevsquincy:
in favour of regular. 

Reason is, we're adding a medium weight of FianacierDisplay, which would be used by any reference to bold FianacierDisplay (it uses the 'nearest' font weight and 'fauxs' it), therefore increasing the number of fonts downloaded on article pages to 4


i-like-robots:
👍

PR #1370, opened 2016-05-09
Splitting out head and main stylesheets

ironsidevsquincy:


andygnewman:
@ironsidevsquincy did you spot any ways in which we could further split the css to reduce the amount that goes in head vs main?

nodemon.json
```diff
@@ -0,0 +1,9 @@
+{
+  "ignoreRoot": [],
+  "watch": [
+    "node_modules",
```
bjfletcher: handy config - how about putting this config in n-makefile, and watching the whole of public folder?

matthew-andrews: +1 i can help with this if people need

ironsidevsquincy: yep, was thinking that. don't need to watch the whole of `public` as only need to restart the server when the `head` css changes (as we read it off the fs to inject into the head)

ironsidevsquincy:
@andygnewman tbh, think you guys would be better placed to decide what goes where, as you know more about the styling. Does it look ok to you? Basic philosophy is anything that's not on the first screen, or styles JS related things, goes in `main`, otherwise `head`

PR #1375, opened 2016-05-11
Check for localStorage value wasn't very good. isNaN returns expected…

seanmtracey:
… value and allows code to run


matthew-andrews:
setting/getting localstorage data directly is a bit dangerous and breaks really badly in safari in private browsing mode.  have you tested it there?  we normally use a [wrapper](https://www.npmjs.com/package/superstore) for our localstorage interaction

seanmtracey:
Didn't know about the wrapper, but makes sense. Will create a PR to amend.

PR #1376, opened 2016-05-11
Add mailing list targeting to the light signup form

tavvy:
Surfaces section data on the light signup component so that users
can be signed up to a specific mailing list. Tweaks the transform
function and adds new tests. Passes new additional data
from article.js via the option object to the transfrom.

Also a few linting fixes


bjfletcher:
LGTM :-)

andygnewman:
Sorry, comments from me though;
I would avoid relying on prefLabel / name fields, unless purely presentational, as these can be changed. However the underlying ID would stay the same. If you need to feed that value to somewhere else consistently or map it back to a search query for results, then ID is the way to go.

PR #1378, opened 2016-05-11
Use id in place of name and sanitise naming

tavvy:
- Use the primarySection id as an identifier instead of its name.
 - Sanitise the naming of the data attributes.
 - Tweak tests to fit


server/transforms/light-sign-up.js
```diff
@@ -6,13 +6,14 @@ module.exports = function ($, flags, options) {
 	const pars = $('p');
 	const variant = flags.lightSignUp;
 	const sectionData = options.metadata.primarySection || null;
-	let mailingList = {
-		name: 'default',
+	let section = {
```
i-like-robots: `const`

andygnewman:
👍

PR #1379, opened 2016-05-11
Update extract-css-block-webpack-plugin to latest version

JakeChampion:
A breaking change was introduced in the previous version. /cc @i-like-robots


i-like-robots:
🚢

Issue #1231, opened 2016-02-08
a11y improvements


i-like-robots:
## Content

- Recommended and More On sections in right hand rail would be better marked up as lists of links rather than a group of headings
- Brand under stand first is confusing as it has no context 

## Widgets

- Social share list needs description. Links need better title (I.E. 'Twitter' -> 'Share this story on Twitter')

GlynnPhillips:
Recommended and More On sections fixed in #1289

PR #1383, opened 2016-05-12
add context to brand information in header

andygnewman:
Resolving second part of [this issue](https://github.com/Financial-Times/next-article/issues/1231).


bjfletcher:
Great, thanks @andygnewman. Speaking of which, our stream page app has some a11y issues that we'll want to fix too.

PR #1391, opened 2016-05-17
using n-ui release candidate to build head and main css

wheresrhys:
@i-like-robots


n-makefile.json
```diff
@@ -2,8 +2,7 @@
   "assets": {
     "entry": {
       "./public/main.js": "./client/main.js",
-      "./public/head.css": "./client/head.scss",
-      "./public/head-ie8.css": "./client/head-ie8.scss",
+      "./public/main.css": "./client/main.scss",
```
i-like-robots: Need a main-ie too =]

i-like-robots: Have pushed https://github.com/i-like-robots/extract-css-block-webpack-plugin/commit/5de39744a0c85ab7477050de930760cd2631ee98 so we can target css extract plugin to particular patterns/files.

i-like-robots: Should simply be:

```scss
$o-grid-mode: 'fixed';
$o-grid-fixed-layout: 'M';

@import 'main';
```

PR #1399, opened 2016-05-23
Use v1 of n-ui

ironsidevsquincy:


wheresrhys:
aaargh nasty. I figure myft-ui will move into n-ui before long so can live with it

PR #1400, opened 2016-05-23
dont extract image from body if its the only thing in it

andygnewman:
@matthew-andrews @GlynnPhillips 

check to see whether there is anything after the image before extracting it.


matthew-andrews:
+!

matthew-andrews:
+1

PR #1408, opened 2016-05-26
add custom paramaters to the light-signup form

tavvy:
We are passing `topic=default` for now in order to check that the values are correctly being handled and passed to the db. Then we will be introducing a dropdown for the user to make a choice.


andygnewman:
LGTM 👍

PR #1414, opened 2016-05-31
removed next-myft-ui templates

wheresrhys:
@keirog @laurieboyes 

FYI - this is the last PR before finally deprecating next-myft-ui


laurieboyes:
woop 👍

PR #1422, opened 2016-06-03
Remove XSLT notes from readme

i-like-robots:
I have added the notes to the [ES interface wiki](https://github.com/Financial-Times/next-es-interface/wiki/XSLT)


andygnewman:
Nice one, had randomly jotted down a note to do this somewhere, sometime, but slipped to back of mind.

PR #1424, opened 2016-06-06
feat(amp): add amphtml link to metadata block

keithamus:
I've hardcoded the amp.ft.com url; it seems to be the convention. If you'd rather I can change it to use a config/env var.


matthew-andrews:
LGTM 👍 

also 👍 for shipping to production on your first day

PR #1427, opened 2016-06-06
feat(amp): add amphtml link to metadata block

keithamus:
I've hardcoded the amp.ft.com url; it seems to be the convention. If you'd rather I can change it to use a config/env var.

(Redo of #1424 but inside FT org so the build works 😄)


matthew-andrews:
nice, does this look OK? @georgecrawford / @quarterto ?

quarterto:
lgtm!

georgecrawford:
me2

keithamus:
Had to rebase, but code is the same.

PR #1431, opened 2016-06-10
Adopt new LSU origami release. Implement collapsible state.

tavvy:
aria compliant, accessible, tab/keyboard friendly. non-breaking, "vanilla" close still supported.
88008472-2d61-11e6-95a3-41fad4f76071


andygnewman:
👍

PR #1432, opened 2016-06-13
implemented skeleton styles for share buttons

wheresrhys:
A little bit of a perf boost, but mainly just a proof of concept to see how easy it is to achieve working with origami components
skeleton share2

@bjfletcher 

@alicebartlett - I like `oShareBase()`, though it goes a little bit beyond what's needed for skeleton styles (e.g link colour). Perhaps the spec could say some thing like 

> The module _may_ implement an `oComponentNameBase()` mixin and if it does the mixin **must** contain all styles which affect dimensions and not cause the download of any additional assets


mmockett:
👍

PR #1433, opened 2016-06-15
using a shared n-ui bundle to initialise js

wheresrhys:
@bjfletcher @ironsidevsquincy @i-like-robots @matthew-andrews


wheresrhys:
This utilises 

- n-express middleware to output nUi minor version https://github.com/Financial-Times/n-express/blob/master/src/middleware/n-ui.js
- n-ui deploy step to keep deploying latest code to https://next.ft.com/n-ui/v1.14/es5.js
- n-ui layout change to output a request for the file https://github.com/Financial-Times/n-ui/blob/master/layout/partials/bootstrapper.html#L34
- one or two tricks in n-webpack to tame the race conditions

I think I should add something to the middleware to verify the n-ui bundle exists before starting the app (e.g. at the moment n-ui/v1.1.5 failed to deploy).

PR #1439, opened 2016-06-20
use n-ui bundle on next-geebee

wheresrhys:
@ironsidevsquincy @matthew-andrews 

Finally ready. Any concerns?


matthew-andrews:
always, but i think let's try it and see

matthew-andrews:
not possible to go behind a flag so we can have quick rollback?

wheresrhys:
I started on that, but it felt like way too much yak shaving (although I
got a reasonable distance before hitting intransigent failing tests in
n-express).

In theory, if we're doing core/enhanced properly* then js breaking for a
bit isn't that big a deal

On Mon, 20 Jun 2016 at 16:38 Matt Andrews <notifications@github.com> wrote:

> not possible to go behind a flag so we can have quick rollback?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/Financial-Times/next-article/pull/1439#issuecomment-227179808>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAbUR0FEnnRdit60I_sUq-to1jHWQrZfks5qNrP5gaJpZM4I50p_>
> .
>

-- 

------------------------------

*This email was sent by a company owned by Financial Times Group Limited 
("FT Group <http://aboutus.ft.com/corporate-information/#axzz3rajCSIAt>"), 
registered office at Number One Southwark Bridge, London SE1 9HL.  
Registered in England and Wales with company number 879531. This e-mail may 
contain confidential information. If you are not the intended recipient, 
please notify the sender immediately, delete all copies and do not 
distribute it further.  It could also contain personal views which are not 
necessarily those of the FT Group.  We may monitor outgoing or 
incoming emails as permitted by law.*

matthew-andrews:
yup, fair enough go go go

PR #1445, opened 2016-06-22
Explicitly turn critical CSS off in IE8

ironsidevsquincy:


bjfletcher:
Thanks @ironsidevsquincy :-) :+1:

PR #1447, opened 2016-06-27
lazy load comments earlier

andygnewman:
@GlynnPhillips 

Trigger loading of comments 600px above the comments, rather than default of 250px.

(On a desktop that would be about 1/2 way through the first 'More-on').


GlynnPhillips:
👍

andygnewman:
Have also added in immediate loading of comments when the request has comment permalink parameters. So will scroll down to that comment.

client/main.js
```diff
@@ -58,13 +58,16 @@ bootstrap(({flags, mainCss}) => {
 			const commentsEl = document.getElementById('comments');
 			const commentsJsLocation = commentsEl.getAttribute('data-comments-js');
 			const commentsCssLocation = commentsEl.getAttribute('data-comments-css');
+			const commentsLazyLoad = commentsEl.getAttribute('data-comments-lazyload') === 'true' ? true : false;
```
bjfletcher: simplify this to:

const commentsLazyLoad = commentsEl.getAttribute('data-comments-lazyload') === 'true';

server/controllers/article.js
```diff
@@ -44,6 +44,8 @@ function isPremiumArticle (webUrl) {
 module.exports = function articleV3Controller(req, res, next, content) {
 	let asyncWorkToDo = [];
 
+	content.lazyLoadComments = (req.query['lf-content'] && req.query.hubRefSrc) ? false : true;
```
bjfletcher: equivalent to:

`content.lazyLoadComments = !req.query['lf-content'] || !req.query.hubRefSrc;`

views/partials/bottom/comments.html
```diff
@@ -3,6 +3,7 @@
 		<div
 			id="comments"
 			data-trackable="comments"
+			data-comments-lazyload="{{lazyLoadComments}}"
```
bjfletcher: `data-comments-lazy-load` (with hyphen in between lazy and load) to be consistent with `lazyLoad` elsewhere.

server/controllers/article.js
andygnewman: Agree more simple, but I find the original way I had it to be a little more readable and understandable of the intent.

client/main.js
andygnewman: Good point.

andygnewman:
Livefyre controls the scroll down - when comments are lazy loaded, the scroll down (and pinpointing of the precise comment) will only occur when the comments have been loaded (ie. when in view, which the user doesn't know).
This PR just overrides the lazy load, so comments are loaded immediately (client side) and then the scroll down occurs - there is still a short time lag, but it at least does what the user expects.

server/controllers/article.js
matthew-andrews: great idea^^

PR #1452, opened 2016-07-02
Don't extract the mainImage from the fragment

### LINKED ISSUE #1451: Article fragment doesn't include master image
This means that if an article that has a master image appears on a stream that has open in stream, when you click open the image disappears.

Noticed it particularly badly when I put [open in stream on photo diary](https://github.com/Financial-Times/next-stream-page/pull/1185) but I've definitely noticed this before on Fast FT too.

See here:-

main-image-disappears

Suggested fix: keep the main image inline with the rest of the content when `fragment=true`.

cc @andygnewman

matthew-andrews:
closes #1451 

main-image-doesnt-disappear

cc @mmockett @andygnewman


andygnewman:
👍

PR #1453, opened 2016-07-04
Redirect FTAlphaville posts to FTAlphaville

matthew-andrews:


andygnewman:
I did just have to check the difference between 301 and 302 - nice correct usage and not a type as I initially thought :)

matthew-andrews:
😄

PR #1458, opened 2016-07-06
Comply with our article grouping convention

commuterjoy:
We have a convention of putting groups of articles in to this HTML hierarchy `section -> article -> a`

### DIFF
--- a/views/partials/related/more-ons.html
+++ b/views/partials/related/more-ons.html
@@ -1,30 +1,32 @@
 {{#each moreOns}}
-<div
-	role="region"
-	data-trackable="more-on-{{@index}}"
-	aria-labelledby="more-on-{{@index}}-title">
-
-	<div class="more-on">
-		<div class="pod">
-			<div class="header">
-				<span class="header__title">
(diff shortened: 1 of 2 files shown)


views/partials/related/more-ons.html
```diff
@@ -1,30 +1,32 @@
...
+<section>
+	<div
+		role="region"
+		data-trackable="more-on-{{@index}}"
+		aria-labelledby="more-on-{{@index}}-title">
+		
+		<div class="more-on">
+			<div class="pod">
+				<div class="header">
+					<span class="header__title">
```
i-like-robots: not your error but this is invalid

keirog: Context: http://w3c.github.io/html/introduction.html#restrictions-on-content-models-and-on-attribute-values

commuterjoy: shall I swap it for a `div` then?

kavanagh:
I don't understand what this is. Should the PR be close or resurrected at some point?

commuterjoy:
yup, old

PR #1459, opened 2016-07-06
Remove duplicate comments tracking

georgecrawford:
**Waiting for confirmation from Lantern before merging this.**

Back in the days of iJento, we subscribed to o-comments events and sent tracking calls to record them. In the meantime, tracking has migrated to Spoor, but o-comments has also implemented its own tracking internally. Next-article therefore sends two events to Spoor: `page:comment` and `comment:post`, for example. This is unnecessary.


tom-parker:
👍

georgecrawford:
OK. Lantern was subscribing to `page:comment` but falling-back to `comment:post`, so this is safe to merge.

PR #1464, opened 2016-07-11
Remove fixed width from gift article notification, fix invalid markup

i-like-robots:
Fixes the issue whereby the fixed width of the gift article message was forcing a horizontal scrollbar on smaller screens. Additionally, this fixes the invalid markup (paragraph as a child of a list) for the inline message variation and ensures the toolbar still has enough space when the comment count goes into triple or quadruple digits.


mmockett:
Hovering over the text itself now seems to show the hand pointer as if it was a link. Is it possible to either remove this state, or actually make it expand / collapse the Gift element?

PR #1478, opened 2016-07-21
Cache 'next up' article

ironsidevsquincy:
V. basic for now, just cache the 'next up' article


andygnewman:
Interesting idea - did you pick the next-up on the basis of data or just an initial presumption that it is the most likely thing to be clicked? Would be interesting to look at the CTR numbers for the article page and see whether it really is? Have a feeling that this article is only clicked on about 1.5% of the time. Will dig out and send you the analysis we did when we last ran some AB tests on links.

ironsidevsquincy:
Yes, it was just an arbitrarily chosen link to start with. 
Going to base it on something like this query - https://beacon.ft.com/data/query-wizard?query=cta%3Aclick-%3Ecount()-%3Egroup(context.domPath)-%3Efilter(context.app%3Darticle)-%3EsortDesc()-%3Eprint(Table) - though need to tidy up the tracking first

PR #1479, opened 2016-07-22
Data trackable attribute tidy

ironsidevsquincy:
Will need to update https://beacon.ft.com/dashboard/articles/ graphs too

@andygnewman


andygnewman:
I notice you removed the word link from the data-trackables. When I was last thinking about the semantics of the data-trackables especially for articles I thought it would be useful to have something that told you what the 'target' or 'objective' of the action was.

I put 'link' in to represent that it was linking to a new piece of content / page. So you could do a query and check whether data-trackable contained link and you'd get all the clicks that went to another piece of content.

Versus clicks that can just expand, save, follow etc.

Wasn't totally thought through, and should probably differentiate between links to content (articles) and links to stream pages.

So overall I think I was aiming for something like 'pagelocation | element-type | target'.

e.g. 'promo-box | headline | article'

Would be nice to try and standardise around something like that.

ironsidevsquincy:
Yeah, I like that. Should it be `content` rather than `article` (to encompass podcasts, etc)? And should there just be a general target for all non-article links, e.g. `action`?

ironsidevsquincy:
Though also, might we be complicating it? Could we just use the referrer to get all links that came from somewhere?

andygnewman:
The issue with referrer is that the data is then related to the target rather than the originator. If we want to look at what type of clicks we are getting on a page we need the data associated with the page the action took place on.

'action' sounds good - may also want 'sharing' for things like email, twitter etc.

ironsidevsquincy:
Say we want to know how many 'content' links are on a page - all we do is a query looking for clicks that have that page's referrer. Is that what you want?

andygnewman:
But putting it in context of the number of views of the original page and compare with number of clicks to other things, in my mind it would seem easier to have it in the CTA collection.

So probably would also want options of topic (eg. stream page):

'content', 'topic', 'action', 'share'

ironsidevsquincy:
Ok, after discussion, in general we might be able to achieve what we want using the referrer. If we find we need to do more complicated queries around differentiating between links to articles and in-page actions, then we might need to add extra data to the data-trackable
Until then though, going to go with this

PR #1481, opened 2016-07-25
Fix brand title styling

leggsimon:
@andygnewman 

The problem was that because the `n-content-tag` class loads its font progressively, its styling has two class selectors and therefore higher specificity than the `article__brand--title` styles we want to override them. This was solved by loading the `article__brand--title` font progressively as well, giving it the same specificity and allowing it to take precedence (as it appears later in the css)

Before:
screen shot 2016-07-25 at 16 20 47

After:
screen shot 2016-07-25 at 16 21 13


matthew-andrews:
:scream: great diagnosis and write up. bit scary it's this fragile though!

andygnewman:
Yikes, agree with @matthew-andrews, should we ensure all typography declarations be progressive then, or just pick them up as and when they cause issues?

matthew-andrews:
i'm not sure, who's our expert on this?  is it doctor @keirog ?

ironsidevsquincy:
I wrote the progressive stuff. Agree, can cause specificity issues, but so far that's always been a good thing as exposed cases where we're not progressively enhancing fonts

leggsimon:
Credit has to go to @laurieboyes for the write up and a lot of the diagnosis though 😃

laurieboyes:
Hooray for teamwork 👬

andygnewman:
👍  on the basis we'll then pick up other issues as and when they occur

PR #1483, opened 2016-07-29
use default and not pointer cursor for the gift banner

bjfletcher:


andygout:
It undoes that undesired cursor behaviour but I can't see what is causing it to act like an interactive element to start with (and maybe it is that which needs to be remedied rather than just overriding it).

bjfletcher:
@andygout originally the gift banner was outside of the button, then I think @i-like-robots moved the gift banner into the button and hence it getting the button pointer.

andygout:
Oh, I see - it's nested with a `li` with a class of `o-share__action` which has a `cursor: pointer` attribute, so it does need to be reversed in this way.

👍

i-like-robots:
I wonder why a non-interactive element (`<li>`) has `'pointer: cursor;` ?

bjfletcher:
@i-like-robots `.o-share__action` assumes the element is interactive, and so it sets the cursor to pointer.  It makes sense, because the gift banner is inside the button, and so it should have this cursor.

I think really the gift banner shouldn't be inside the button, but it's already been done and I think it's not worth the effort to rework all that again. That is especially so given that all this code will be sent to `/dev/null` once the banner has served its time. :)

PR #1484, opened 2016-08-02
feat: add basic start for anon sample articles

keithamus:


server/controllers/article-helpers/sample-articles-data.js
```diff
@@ -0,0 +1,318 @@
+module.exports = [
+	{
```
keithamus: This is just a bunch of hardcoded content, because the list is curated. Ideas on how to better do this would be welcome 😄

client/components/sample-articles/_main.scss
```diff
@@ -0,0 +1,122 @@
...
+	max-width: 736px;
+	background: #316f9c;
+	margin: 0 48px 48px;
+	color: white;
+
+	@media (min-width: 760px) {
+		margin: 0 auto;
+	}
+
+	&__title {
```
ironsidevsquincy: I'd try and avoid this pattern, makes it hard to search

Also, does it need to be nested in `.sample-articles`? Isn't `.sample-articles__title` specific enough?

client/main.js
```diff
@@ -3,6 +3,7 @@
 const oViewport = require('n-ui/viewport');
 const nVideo = require('n-video');
 const lightSignup = require('o-email-only-signup');
+const expander = require('o-expander');
```
wheresrhys: Could you `require('n-ui/expander')` (which just wraps o-expander), then you won't need to add it in your bower.json either.

This probably seems weird, but is due to historical problems with consistency and conflicts when using components across our services.

I may attempt to add some linting for this sort of thing somehow. Any ideas welcome

client/components/sample-articles/_main.scss
```diff
@@ -0,0 +1,122 @@
+.sample-articles {
+	display: flex;
+	flex: 1 1 auto;
+	flex-wrap: wrap;
+	max-width: 736px;
+	background: #316f9c;
```
ironsidevsquincy: Best to use the colours from Origami, if you can http://registry.origami.ft.com/components/o-colors

```background: getColor('pink');```

File: client/components/sample-articles/_main.scss [hunk 5]
```diff
@@ -0,0 +1,122 @@
...
+	margin: 0 48px 48px;
+	color: white;
+
+	@media (min-width: 760px) {
+		margin: 0 auto;
+	}
```
ironsidevsquincy: Can use `oGridRespondTo` mixin for responding to breakpoints - https://github.com/Financial-Times/o-grid#responsive-layout-helper

client/components/sample-articles/_main.scss
```diff
@@ -0,0 +1,122 @@
...
+
+	}
+
+	&__more-button {
+		@include oTypographySans(s);
+		color: white;
+		padding: 8px 0 16px;
+		margin: 0;
+
+		i {
```
ironsidevsquincy: can we get a class on this, rather than using the element?

client/components/sample-articles/_main.scss
keithamus: The nesting just expands selectors. The compiled output only has one class; `.sample-articles__title`. I can expand them all though if you'd prefer.

client/components/sample-articles/_main.scss
```diff
@@ -0,0 +1,122 @@
...
+		max-width: 366px;
+		border: 0;
+	}
+
+	&__wrapper__lead {
+		margin: 0 24px;
+	}
+
+	&__header {
+		@include oTypographySans(xs);
```
ironsidevsquincy: these take a second parameter, `$load-progressively`, that loads the fonts progressively - https://github.com/Financial-Times/o-typography#progressively-loading-web-fonts

server/controllers/article-helpers/sample-articles.js
```diff
@@ -0,0 +1,90 @@
...
+		map((topic) => indexOf(contentId)(topic.articles) !== -1)(sampleArticles)
+	);
+}
+
+module.exports = function getCuratedArticles (contentId) {
+	const currentTopicIndex = topicIndexOf(contentId);
+	const articlesWithoutCurrent = map((topic) => ({
+		title: topic.title,
+		articles: without(contentId)(topic.articles)
+	}))(sampleArticles);
```
ironsidevsquincy: any reason not to use `Array`'s `map`?

client/components/sample-articles/_main.scss
wheresrhys: :+1: for not using sass's BEM concatenation. It does indeed make finding code very difficult

ironsidevsquincy: Ah yeah, ignore the specificity bit. Still makes it hard to search though

client/components/sample-articles/_main.scss
keithamus: sadly no. This is generated by o-expander, and it doesn't provide a class for it.

server/controllers/article-helpers/sample-articles.js
keithamus: Not especially just keeping consistent as I need to use it on L72

server/controllers/article-helpers/sample-articles-data.js
ironsidevsquincy: Bit late, but we have an api called Bertha, which reads in a Google spreadsheet and spits out something json formatted

ironsidevsquincy: https://github.com/Financial-Times/next-article/blob/3b1f06d8ec304e1e2fae4452d4fea076b24528fa/server/lib/ig-poller.js

server/controllers/article-helpers/sample-articles.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+const api = require('next-ft-api-client');
+const get = require('lodash/fp/get');
+const map = require('lodash/fp/map');
+const flatten = require('lodash/fp/flatten');
+const without = require('lodash/fp/without');
+const indexOf = require('lodash/fp/indexof');
```
ironsidevsquincy: `const indexOf = require('lodash/fp/indexOf');`?

keithamus: 😞 yes I've just worked this out... same as `keyby`

PR #1493, opened 2016-08-09
Remove unused ads flag

adgad:
/cc @andrewgeorgiou1981 @VladDubrovskis 

Luckily, this would have broken if the flag expired as that hlfmpu template doesn't exist 🙈


VladDubrovskis:
👍

PR #1498, opened 2016-08-10
Remove lazy loading from slots so they can be controlled centrally

adgad:
/cc @VladDubrovskis @andrewgeorgiou1981


VladDubrovskis:
👍 LGTM, run a rebuild as don't see a reason for build to have failed with this change

PR #1508, opened 2016-08-12
Check for comment fragment indicator before lazy loading comments

andygnewman:
Building on previous fix to stop lazy loading comments when customers using a permalink to a comment, this aims to stop lazy loading when the deeplinking to a comment is done through a fragment indicator rather than a query param.

This type of URL is generated by livefyre for moderation of comments.


client/components/comments/lazy-load.js
```diff
@@ -27,7 +27,8 @@ export default opts =>
 	new Promise((resolve, reject) => {
 		const target = document.querySelector(opts.targetEl);
 		if (target) {
-			if (opts.commentsLazyLoad && window.IntersectionObserver) {
+			const commentsFragmentIndicator = /lf-content/.test(window.location.hash);
```
ironsidevsquincy: This counts as a micro-performance tweak, so not worth doing, but when it comes to string searching, regexs are the slowest thing you can use

In this case, as you're not using anything specific to regex (anchors, character classes, etc), you might as well use `window.location.hash.indexOf('#lf-content') > -1` (assuming the fragment starts with that string)

ironsidevsquincy:
👍

PR #1512, opened 2016-08-15
fix encoding issues

bjfletcher:
A fix has been implemented with `n-heroku-tools` and so here we're just going to use that with a "dummy" JS change so that we have a new hashed asset hopefully with correct encoding 👍


client/components/share/main.js
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { EmailArticleData, emailArticleModes, EmailArticleView } from 'n-email-article';
-
-const OShare = require('o-share');
+import OShare from 'o-share';
```
wheresrhys: :+1: (let's hope babel doesn't convert them both to the same thing)

PR #1514, opened 2016-08-17
Aud Dev users see inferred variant of LSU

ironsidevsquincy:


server/controllers/article.js
```diff
@@ -43,6 +43,8 @@ function isPremiumArticle (webUrl) {
 	return webUrl.search('/cms/s/3') !== -1
 }
 
+const isAudDev = req => req.query.exe === '16q3beacon';
```
ironsidevsquincy: Useful for you @keithamus?

PR #1516, opened 2016-08-17
feat(anonSampleArticles): only activate if isAudDev. Drop isAudDev co…

keithamus:
@ironsidevsquincy I've changed isAudDev to drop a cookie, as we need sample articles to persist accross all sample articles. Hopefully this is okay for LSU requirements?


server/controllers/article.js
```diff
@@ -43,7 +43,19 @@ function isPremiumArticle (webUrl) {
 	return webUrl.search('/cms/s/3') !== -1
 }
 
-const isAudDev = req => req.query.exe === '16q3beacon';
+function isAudDev (req, res) {
+	if (/[; ]exe=16q3beacon[; ]/.test(req.headers.cookie)) {
```
ironsidevsquincy: Maybe worth using a cookie parser middleware - https://www.npmjs.com/package/cookie-parser

ironsidevsquincy:
Yep, that's better for LSU actually

👍

PR #1521, opened 2016-08-20
feat: Switch tag links to vanity URLs

matthew-andrews:
@andygnewman @andygout @bjfletcher


server/mappings/article-pod-mapping-v3.js
```diff
@@ -37,7 +37,7 @@ module.exports = function articlePodMapping(article) {
 
 	if (decoration.primaryTag) {
 		Object.assign(decoration.primaryTag, {
-			url: `/stream/${decoration.primaryTag.taxonomy}Id/${decoration.primaryTag.idV1}`
+			url: decoration.primaryTag.url.replace('https://www.ft.com', '')
```
matthew-andrews: This actually doesn't do anything anymore because `decoration.primaryTag` has a `url` property already on it (although it's wrong - it contains the `https://www.ft.com` prefix).

You can see this on the live site here:-
https://www.ft.com/content/90ce80bc-6606-11e6-8310-ecf0bddad227

Look at the URL when you hover over the tag under ‘More on this topic’ …

matthew-andrews:
Hm, just discovered there are a few podcasts that haven't re-ingested.  Don't merge until I do that :)

i-like-robots:
https://github.com/Financial-Times/next-es-interface/wiki/Podcasts#run-a-complete-ingest-immediately 👀

matthew-andrews:
yah that's fine but i think i have forgotten to implement the vanity lookup for podcasts 😁

PR #1524, opened 2016-08-23
test(anonSampleArticles): fix failing tests due to header change

keithamus:
@bjfletcher 😓


server/controllers/article-helpers/decorate-metadata.js
```diff
@@ -7,7 +7,7 @@ function fillProperties(article) {
 		let v1 = {
 			id: tag.idV1,
 			name: tag.prefLabel,
-			url: tag.url.replace('https://www.ft.com', '')
+			url: String(tag.url || '').replace('https://www.ft.com', '')
```
keithamus: Just added some defensive guards here because this was failing on occasional tests, due to url sometimes being undefined.

PR #1525, opened 2016-08-23
feat(anonSampleArticles): add data-tracking attrs to blocks

keithamus:
This adds a few data-tracking attrs to various nodes to assist in the tracking of clicks in the Sample Articles component

### DIFF
--- a/views/partials/related/sample-articles.html
+++ b/views/partials/related/sample-articles.html
@@ -1,4 +1,4 @@
-<section class="sample-articles">
+<section class="sample-articles" data-tracking="sample-articles">
 	<header class="sample-articles__ribbon">
 		<div class="sample-articles__ribbon__wrapper jump-to-block" id="more-sample-articles">
 			<h3 class="sample-articles__ribbon__heading">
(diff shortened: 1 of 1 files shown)


views/partials/related/sample-articles.html
```diff
@@ -1,4 +1,4 @@
-<section class="sample-articles">
+<section class="sample-articles" data-tracking="sample-articles">
```
ironsidevsquincy: attribute is `data-trackable`

views/partials/related/sample-articles.html
```diff
@@ -42,7 +43,7 @@ <h4 class="sample-articles__item__headline">
 	{{/with}}
 		<div class="sample-articles__wrapper">
 			<h3 class="sample-articles__header__lead">Recommended Reads</h3>
-			<ul class="sample-articles__items">
+			<ul class="sample-articles__items" data-tracking="sample-articles-other-topics">
```
ironsidevsquincy: no need to duplicate the `sample-articles-` bit, the tracking recurses up the dom, pulling out all the `data-trackable` values, to create the full tracking id, e.g. in this case it would be

```
...sample-articles | sample-articles-other-topics
```

so you'll just want it to be `data-trackable="other-topics"` here so it becomes

```
...sample-articles | other-topics
```

ironsidevsquincy: https://github.com/Financial-Times/n-ui/blob/9bbf3678182ffa997453de394819c87c63f572c2/tracking/test/ft/utils/getDomPathString.spec.js

keithamus: cool I'll update 👍

PR #1527, opened 2016-08-23
use NPM instead of Bower for the JS

bjfletcher:
of `n-email-article`


package.json
```diff
@@ -30,6 +30,7 @@
     "shellpromise": "^1.3.0"
   },
   "devDependencies": {
+    "@financial-times/n-email-article": "^1.2.8",
```
ironsidevsquincy: shouldn't this be in `dependencies`?

bjfletcher: :+1:

PR #1528, opened 2016-08-23
Vary articles the same as product selector

adgad:
This is an attempt to fix a bug where FCF articles still show a cached article (instead of a barrier) when opened in a different tab without the referer. It adds the cache keys from https://github.com/Financial-Times/next-product-selector/blob/master/server/middleware/vary.js

/cc @GlynnPhillips @commuterjoy @phamann


wheresrhys:
Can we add a test so we don't get caught out by this in future? e.g in article ping a prod barrier to read the vary header and check it's the same as what article sets, and vice versa

wheresrhys:
image
grumman?!

adgad:
Ha! I like to keep myself a reminder of the good old days.

Yup test would be a good idea. Want to verify that this actually fixes the problem first.

PR #1534, opened 2016-08-26
final stab at RUM solution for comments for all users

wheresrhys:


i-like-robots:
What's the 1.6mb file?

PR #1536, opened 2016-08-26
Split authors out from the tags in the myFT mentions area

keirog:
Before:
before

After:
after


views/partials/related/more-ons-tags.html
```diff
@@ -31,6 +32,30 @@ <h3 class="more-ons-tags__title-text">
 		</li>
 		{{/each}}
 	</ul>
+	{{/if}}
+	{{#if authors}}
+	<div class="n-content-tags__intro">
+		{{#ifEquals contentType 'podcast'}}
+			<p>Follow the authors related to this show</p>
```
laurieboyes: Whoa, psychic-level issue spotting powers ✨

laurieboyes: I wait it's the same in topics lol

views/partials/related/more-ons-tags.html
```diff
@@ -31,6 +32,30 @@ <h3 class="more-ons-tags__title-text">
...
+		{{else}}
+			<p>Follow the authors of this {{contentType}}</p>
+		{{/ifEquals}}
+	</div>
+	<ul class="n-content-tags__list--myft">
+		{{#each authors}}
+		<li class="n-content-tag__list-item--with-follow">
+			<a
+				href="{{url}}"
+				data-trackable="topic"
```
laurieboyes: Should we change the trackable value to `author`?

laurieboyes: Also, don't know if it's necessary or not and aware that this is the same as for topic, but does this allow us to identify which author was clicked in beacon?

laurieboyes:
👍 apart from probably non-blocking comments

views/partials/related/more-ons-tags.html
andygnewman: Good idea to change trackable to author.
We did institute indexing of links for more-ons in articles and for article cards in streams to help identify the right number to show based on interactions, if that's the kind of use case you were thinking of.
https://github.com/Financial-Times/next-article/blob/master/views/partials/related/pod-item.html#L2

laurieboyes: Ah I was thinking passing the conceptId for the author as part of the tracking data, rather than its index in the list

laurieboyes: Oh wait! We'll already have that for the follow buttons. The links are much less important. Cool, ignore me 👍

andygnewman: I thought you would be tracking specific author following from the backend, but I'm guessing you don't have a source identifier on a follow action then. Would be useful to see where different concepts get followed from.

keirog: Oops thanks all, missed that trackable bit. That should be enough for what we need 👍

PR #1537, opened 2016-08-26
premium labelling

bjfletcher:
using Ammit


client/components/premium-label/main.js
```diff
@@ -0,0 +1,34 @@
...
+	// pretty raw :/
+	const premiumLabel = document.createElement('span');
+	premiumLabel.classList.add('article__premium-label');
+	premiumLabel.appendChild(document.createTextNode('Premium'));
+	const articleTitle = article.querySelector('.article__title');
+	articleTitle.appendChild(premiumLabel);
+};
+
+const isFeatureEnabled = () => {
+	return fetch('https://ammit-api.ft.com/allocate', {
```
bjfletcher: this doesn't work locally because of CORs
how about www.ft.com/ammit-api?

client/components/premium-label/main.js
```diff
@@ -0,0 +1,34 @@
...
+			headers: {
+				'ft-session-token': session.cookie()
+			}
+	}).then(() => true); // question: how to work it out?
```
bjfletcher: @tom-parker can you confirm how we'd work out whether the feature is enabled here?

bjfletcher:
image

client/components/premium-label/main.js
matthew-andrews: doesn't ammit simply set flags?

wheresrhys: @tom-parker @bjfletcher you can add ammit to here https://github.com/Financial-Times/next-router/blob/master/server/models/api.js and add `useCorsProxy: true` to teh fetch request and it _should_ just work

wheresrhys:
looks good. The only worry I have is that if the premium label would cause the text to run on to another line, causing content to jump around. But having said that, myft buttons have the same problem, so not _that_ big a deal

client/components/premium-label/main.js
tom-parker: You should be able to do something like 
```
}).then((resp) => console.log(resp.headers.get('x-ft-ab'));
```

client/components/premium-label/main.js
```diff
@@ -0,0 +1,34 @@
+const session = require('ft-next-session-client');
+
+const isArticlePremium = (article) => {
+	// the server adds is-premium attribute to article if it is premium
+	return article.hasAttribute('data-is-premium');
+};
+
+const addPremiumLabel = (article) => {
```
andygnewman: Given we'll have to put in a class server side, we may as well put in the span element then aswell, then all we have to do client side is put a class in to switch on the styling.

client/components/premium-label/main.js
bjfletcher: @matthew-andrews by reading cookies client-side for the flag, you mean?

client/components/premium-label/main.js
bjfletcher: @andygnewman either options (method you explained and method used in this PR) are fine with me... I just thought that it was clearer to have everything under `client/components/premium-label/` with the CSS in `_main.scss` and the JS and HTML in `main.js`.  All in the same one folder.  I think both methods have pro's and con's and so I'm open minded here...

bjfletcher:
@wheresrhys regarding premium label causing a reflow, what are the options here? I can think of:
1. place label somewhere that can't cause a line break
2. have a placeholder (although if premium label isn't added then there'll be some obvious gap)
3. have "Standard", "Free" as well as "Premium" labelling, all with the same width.

PR #1539, opened 2016-08-30
Redirect syndicated content to old ft.com

andygnewman:
When content is not originating from the FT (eg. Reuters), redirect back to old ft.com.
Temporary measure until agreed / implemented what syndicated content on next should look like.

Should be implemented before wires content is ingested through this [PR](https://github.com/Financial-Times/next-es-interface/pull/482)


matthew-andrews:
+1

PR #1540, opened 2016-08-30
premium labelling

bjfletcher:
This is the server-side rendering alternative. I think quite straightforward to extract when/if we change how we do things.

/cc @wheresrhys @andygnewman


client/components/premium-label/_main.scss
```diff
@@ -0,0 +1,13 @@
+.article__premium-label {
+	@include oTypographySansBold('xs');
+	margin-left: 10px;
+	vertical-align: middle;
+	border-radius: 3px;
+	background-color: #800080;
```
andygnewman: I think that the background color suggested on the trello cards for Premium is more of a grey than a purple - you might want to check with Ollie. Try and persuade the designers to use colors already in the FT color palette http://registry.origami.ft.com/components/o-colors.

client/components/premium-label/_main.scss
```diff
@@ -0,0 +1,13 @@
...
+	color: #ffffff;
+	padding: 2px 5px;
+
+	@include oGridRespondTo(L) {
+		@include oTypographySansBold('s');
```
andygnewman: you can use the size / line height only mixin here `oTypographySansBoldSize('s')`

andygnewman:
Just a couple of CSS comments, but otherwise 👍

client/components/premium-label/_main.scss
ironsidevsquincy: Also, the progressive font version of the mixin - https://github.com/Financial-Times/o-typography#progressively-loading-web-fonts

bjfletcher: looks like size/line-height only mixin doesn't support progressive font loading:

```
    ERROR in ./client/main.scss
    Module build failed:
       		@include oTypographySansBoldSize(s, $load-progressively: true);
          Mixin oTypographySansBoldSize has no parameter named $load-progressively
```

if I omit the load progressively parameter here, keeping it just in the mobile one, then mobile's progressively loaded font CSS overrides desktop's because of it's greater specificity.

ironsidevsquincy: Sorry, should have been more detailed, only put it in the default layout's font. In `L` layout, use the `Size` mixin, e.g.

```
.article__premium-label {
    @include oTypographySansBold('xs', $load-progressively: true);
    ....
    @include oGridRespondTo(L) {
        @include oTypographySansBoldSize('s');
    }
}
````

Comment (bjfletcher) [inline client/components/premium-label/_main.scss]: @ironsidevsquincy the specificity issue I mentioned is this:
image

You see how the progressive loading CSS for mobile overrides the sizing CSS for desktop?

In an attempt to racify that, I added progressive loading parameter to the size one too... but this didn't work either because of the error I mentioned earlier (size doesn't allow the parameter).

Comment (bjfletcher) [inline client/components/premium-label/_main.scss]: Another issue I have is this.  Note how the premium label moves around?  Presumably because font sizes for font used before FT font is loaded, and the FT font, don't match up.

font-issue-2

Comment (ironsidevsquincy) [inline client/components/premium-label/_main.scss]: re: specificity, yes, that's expected, the web font is (annoyingly) a different size to the fallback font.

The moving font is because we drop a cookie (`next-fonts-loaded`) once the fonts load the first time, which means the progressive font classes are applied immediately (https://github.com/Financial-Times/n-ui/blob/master/typography/enhance-fonts.html). However if you do a hard refresh, the cookie is retained but the fonts are re-downloaded, hence the 'flash' of wrong-sized fonts. If you perform a standard user-journey (e.g, coming to this page from a link), this won't happen

Comment (ironsidevsquincy) [inline client/components/premium-label/_main.scss]: Oh, in my above snippet, missed out

```
@include oTypographySansBoldSize('s', $with-progressive-size: true);
```

(as per the linked to Readme)

PR #1542, opened 2016-08-31
Add placeholder title to videos in articles

andygnewman:
cc: @ironsidevsquincy 

To solve the issue raised here - https://financialtimes.slack.com/archives/ft-next-support/p1472625414000060


ironsidevsquincy:
👍  (weird they've gone for all lowercase for those configuration properties...)

PR #1543, opened 2016-08-31
Additional article MPU slots

adgad:
@VladDubrovskis @andrewgeorgiou1981 @andygnewman 
If the adsMoreArticleMPUs flag is on:

* If an article is shorter than 3 paragraphs, stick an ad at the bottom of the article
* On mobiles and tablet portrait - add a second ad slot a further 6 paragraphs after the first one (still abiding to the same rules of having to be between two paragraphs and not an aside, etc).

Also tidy up some old classnames.

### DIFF
--- a/server/transforms/inline-ad.js
+++ b/server/transforms/inline-ad.js
@@ -1,12 +1,10 @@
-module.exports = function ($) {
+'use strict';
+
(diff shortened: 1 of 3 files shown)


VladDubrovskis:
👍

PR #1550, opened 2016-09-01
More ons use n card

andygnewman:
Get the more-ons to use n-card rather than own bespoke template so can inherit all the n-card goodness.

So would be rendered like this
screen shot 2016-09-01 at 18 02 31

Rather than like this currently
screen shot 2016-09-01 at 18 03 11


Procfile
```diff
@@ -1 +1 @@
-web: forever -c 'node --harmony' server/app.js
+web: forever -c 'node --harmony' server/init.js
```
bjfletcher: This may be a bit confusing - it's not some "init" code that we run... it's an actual, full app that we're running here.

server/init.js
```diff
@@ -0,0 +1,2 @@
+require('./setup');
+module.exports = require('./app');
```
bjfletcher: Any particular reason for not putting `require('./setup');` at the top of the `app.js`?

bjfletcher:
Awesome work @andygnewman

All looking really good apart from the naming of `init.js` in `Procfile`.

server/init.js
ironsidevsquincy: Every `require` after the setup gets babel-ified; means `app.js` can use ES6 features

andygnewman:
I was following the pattern established by the front-page for doing this. https://github.com/Financial-Times/next-front-page/pull/1012 by @keirog / @ironsidevsquincy 

Makes sense to have the naming that way round in establishing things first (such as babelification) before then starting the actual app, meaning that the entry point is the init.

bjfletcher:
It feels very strange to call `init` to start the app. Any app has all sorts of init code to do. Maybe a better name would be `server/app-babel.js`?

ironsidevsquincy:
What in particular is strange about called it `init`? The file contains a bit of 'initialisation' work that needs to be run (can potentially contain anything, not just setting up babel), before running the main app, so seems to make sense to me...

keirog:
Come to think of it, it does feel odd in terms of it breaking the general convention. Maybe a `require('./babelify');` or similar at the top of `app.js` is clear enough? Not sure if it was kept separate for any technical reason. (Testing?)

ironsidevsquincy:
Setup was in a separate file so it could be used by tests (https://github.com/Financial-Times/next-front-page/blob/master/Makefile#L7)

And as mentioned above (https://github.com/Financial-Times/next-article/pull/1550#discussion_r77334823), can't add the babel-register at the top of `app.js` if you want `app.js` to be babelified

PR #1554, opened 2016-09-02
Story package use n card

andygnewman:
Not necessarily the most attractive rendering, but this is 'straight out the box n-card' without any further design. Suggest we get Ollie to take a look and see what small tweaks, we can do, consistent with other cards, to make it a little prettier.

screen shot 2016-09-05 at 08 20 17


andygout:
Nice. What happens if only two or four Related Stories are available - do you get an empty slot on the right (although I have a feeling I've missed whatever prevents this)?

andygnewman:
You're right, there's no code to make it to an odd number of articles.
Wasn't in the original code base, so haven't gone backwards, but you're right it should be there.

Will do a quick amend.

PR #1564, opened 2016-09-06
Slot in related content for promoted articles

adgad:
@VladDubrovskis @gvonkoss @andygnewman @bjfletcher 
Adds a slot at the bottom of an article (should be between moreons) that shows either a paid post, special report(s) or Smartmatch native content ( behind a flag )

- [x] Slot for paid posts/special reports
- [x] Get second article if first is a special report
- [x] Hide for Core Experience / if ad doesnt load
- [x] New feature flag
- [x] Smartmatch
- [ ] Smartmatch "external content" icon
- [x] Tracking
- [x] Move between moreons (note: double check moreon tracking doesn't break)
- [ ] Check it works with real ads
- [ ] Check cross browser - Smartmatch ot working in IE, but looks like it's a bug with the response they give.

screen shot 2016-09-08 at 12 52 00


VladDubrovskis:
👍

PR #1566, opened 2016-09-07
simpler allSettled implementation

wheresrhys:


client/components/onward-journey/main.js
```diff
@@ -85,7 +75,9 @@ module.exports.init = () => {
 		);
 	}
 
-	return allSettled(fetchPromises)
+	return Promise.all(
+		fetchPromises.map(p => p.catch(() => null))
+	)
 		.then(() => {
 			lazyLoadImages();
```
andygnewman: Given this is client side I'm not clear why we're adding lazy loading on as well. Will do another PR to clean that up.

andygnewman:
LGTM

client/components/onward-journey/main.js
wheresrhys: that's funny as the original intent of this PR was to remove that (replacing with something from n-ui) but it _is_ needed

ironsidevsquincy: yeah, to lazy load the images in the more ons

PR #1568, opened 2016-09-08
abstract the premium label

bjfletcher:
I've extracted the premium label so that can easily be used elsewhere on the article in other components.

/cc @andygout


andygout:
👍 

Surprised that `.article-headline__title` was not being used by itself to style anything.

PR #1569, opened 2016-09-08
Add premium label to some other article components

bjfletcher:
- the read next components
- the suggested reads component

The asides component is in XSLT and we probably won't do it.

Note that this branches off https://github.com/Financial-Times/next-article/pull/1568


client/components/read-next/main.scss
```diff
@@ -67,6 +68,14 @@
 
 }
 
+.next-up__headline-text {
+	color: oColorsGetPaletteColor('cold-2');
+
+	.next-up__headline:hover & {
+		border-bottom: 1px dotted;
```
andygout: I wonder if there is an include for this somewhere given it is the standard hover styling.

client/components/suggested-reads/main.scss
```diff
@@ -39,10 +39,20 @@
...
+.suggested-reads__list-item__headline-text {
+	display: inline;
+	color: oColorsGetPaletteColor('cold-2');
+	font-size: inherit; // nullify H3's override
+	font-weight: inherit; // nullify H3's override
 
-	a {
-		@include nLinksHeadline();
+	.suggested-reads__list-item__headline:hover & {
+		border-bottom: 1px dotted;
```
andygout: Ditto.

andygout:
One note (which you've probably already considered), but otherwise 👍

client/components/read-next/main.scss
bjfletcher: https://github.com/Financial-Times/next-sass-setup/blob/63720f227fc6bcfda9a33df1e89cb88c35d71da0/_links.scss#L22

Problem with that is we don't want the dotted border with the premium label, hence this custom one.

views/partials/header/read-next-header.html
```diff
@@ -3,7 +3,14 @@
 		<div class="next-up next-up__header" data-trackable="next-up | top">
 			<h3 class="next-up__intro next-up__intro__header">{{#if moreRecent}}Read latest:{{else}}Read next:{{/if}}</h3>
 			<article>
```
i-like-robots: not a part of this PR I realise but this is not a valid use of the article element (this content cannot be standalone)

i-like-robots: whoops, too late =]

PR #1570, opened 2016-09-08
run "read more" & "suggested reads" through n-content-model

bjfletcher:
/cc @andygout @andygnewman


test/server/controllers/article-helpers/read-next.test.js
```diff
@@ -149,4 +145,23 @@ describe('Read next', function() {
...
+				Promise.resolve({
+					id: '9a2b7608-5746-11e5-9846-de406ccb37f2',
+					webUrl: 'http://www.ft.com/cms/s/3/123456'
+				})
+			);
+		});
+
+		it('should be true', () => {
+			return subject('', [ '' ], {}, '2015-09-10T18:32:34.000Z').then(result => {
+				result.premium.should.be.truthy;
```
andygout: Is this saying that `result.premium` attribute should exist rather than its actual value being `true`?

test/server/controllers/article-helpers/suggested.test.js
```diff
@@ -131,4 +131,21 @@ describe('Suggested reads', () => {
...
+		before(() => {
+			stubs.content.returns(Promise.resolve([{
+				id: '9a2b7608-5746-11e5-9846-de406ccb37f2',
+				webUrl: 'http://www.ft.com/cms/s/3/123456'
+			}]));
+		});
+
+		it('should be true', function() {
+			return subject(articleId, []).then(results => {
+				expect(results[0].premium).to.be.truthy;
```
andygout: Ditto.

server/controllers/article-helpers/suggested.js
```diff
@@ -32,6 +31,7 @@ module.exports = function(articleId, storyPackageIds, primaryTag) {
 				index: 'v3_api_v2'
 			})
 		)
+		.then(items => items.map(i => contentModel(i, { useCase: 'article-card', excludeTaxonomies: true })))
```
andygout: `item` instead of `i` for readability?

andygout:
In the tests what is setting the `premium` value to true? is it the default behaviour of calling on a stub? (Am asking for my own education).

bjfletcher:
@andygnewman It uses the actual `n-content-model` library, rather than stubbing it out.

This library adds the `premium` property.

We could stub this library out but then we have no test that app <-> `n-content-model` is working ok.

There are pro's and con's for either approach.

andygout:
So if tests fail it could be the compatibility between the app and `n-content-model` or just the app by itself.

I see the use in testing the compatibility but shouldn't they distinguish between the two? Is the solution maybe to write more tests - one set for each scenario?

bjfletcher:
@andygout http://martinfowler.com/articles/microservice-testing/

one extreme is stubbing out absolutely everything outside of the module - e.g., if we use `jQuery.each` then stub it out. :)

the other extreme is to stub out absolutely nothing - everything is real... more like an integration test

my view is that because `n-content-model` doesn't use HTTP and is just some transformer, then we needn't stub it out.

If we bumped version of `n-content-model` to a new major version, how'd we know to check that the `premium` attribute is still set?  `js-success` smoke test/nightwatch won't tell us.

andygnewman:
I'm in general a 'stubber' and test for whether a service is being called, but not the value returned by the stubbed module, but can see the impact of changes to the stubbed module not being easily caught.

In general it's not something we have a clear strategy on and I'm reasonably unbothered by having a couple of tests that use a value returned from an independent module, but that doesn't look to cover all the scenarios that the module may have to cope with.

PR #1573, opened 2016-09-13
Amend: Classification class to classifier class

andygout:
@bjfletcher 

Some outstanding instances that need changing in relation to this [`n-card` PR](https://github.com/Financial-Times/n-card/pull/183/files).


bjfletcher:
I thought I had done this, obviously I didn't push or merge :/ Definitely a :+1: from me. Nice one Andy :)

bjfletcher:
Wait....

bjfletcher:
Sorry, could you please include it in critical CSS? As suggested reads, etc., are above the fold.

(BTW, just so you're aware, we've got this separate style for the article headline: https://github.com/Financial-Times/next-article/blob/master/client/components/classifier/_main.scss)

andygout:
^^^ I _think_ this is the required fix.

Also, there is some disparity in the BEM logic used to name the classifier gaps both here in [`next-article`](https://github.com/Financial-Times/next-article/blob/master/client/components/classifier/_main.scss) and in [`n-card`](https://github.com/Financial-Times/n-card/blob/86a6b134a264395ce6143359eaa8ee5b34578a5c/components/_classifier.scss). I'd argue that `classifier-gap` is a modifier of the `article` or `card` element respectively and that both should adhere to the model that `n-card` uses. What do you think (I can make changes once we've decided)?

client/main.scss
```diff
@@ -12,6 +12,7 @@ $n-ui-output-critical: $output-critical;
 
 @include nUiCriticalStart();
 	@import 'n-image/main';
+	@import 'n-card/main';
```
bjfletcher: `@import 'n-card/components/classifier';`

(we're using just that bit above the fold)

andygout: I was wondering if we could be that specific. Nice that we can.

bjfletcher:
:+1:

views/partials/right-hand/suggested-reads.html
```diff
@@ -15,7 +15,7 @@ <h3 id="suggested-reads__title" class="suggested-reads__title">
 			{{#each articles}}
 				<article class="suggested-reads__list-item" data-trackable="suggested-read | {{@index}}">
 					<a href="{{url}}" class="suggested-reads__list-item__headline" data-trackable="headline">
-						<span class="card__classification-gap">
+						<span class="card__classifier-gap">
 							<h4 class="suggested-reads__list-item__headline-text">{{title}}</h4>
```
i-like-robots: again not from this PR but this HTML is invalid (spans are phrasing content - phrasing content cannot contain flow content =]

PR #1576, opened 2016-09-13
Sticky-right-rail ad MVT - ad moved to below the 'more on' section

andrewgeorgiou1981:


views/partials/right-hand/right-hand.html
```diff
@@ -1,10 +1,22 @@
...
-	{{> ads/mid}}
+	{{#if @root.flags.stickyRightAd}}
 
-	{{{tocHTML}}}
+		{{#if @root.flags.articleSuggestedRead}}
+			{{> right-hand/suggested-reads articles=readNextArticles topic=readNextTopic }}
+		{{/if}}
+
+		{{> ads/mid}}
+		{{{tocHTML}}}
```
adgad: I think this need to be above the ad as well

PR #1578, opened 2016-09-13
Update canonical url for SEO love

richard-still-ft:


server/controllers/article.js
```diff
@@ -160,6 +168,8 @@ module.exports = function articleV3Controller(req, res, next, content) {
 		isInferred: res.locals.flags.lsuInferredTopic
 	};
 
+	content.canonicalUrl = getCanonicalUrl(content.webUrl, content.id);
```
matthew-andrews: fussy but would it be ok if we moved this to the line before `content.layout = 'wrapper';` — that's the only case this is needed.

richard-still-ft: Should be bonafide now @matthew-andrews

matthew-andrews:
oh we're doing this now?  crazy times.

PR #1580, opened 2016-09-14
Linting

andygout:
cc @andygnewman @bjfletcher


bjfletcher:
Wow, no `eslint` warnings whatsoever now. 🎉 🎉 🎉 nice work @andygnewman ! :)
Should we make the warnings errors now? So we don't let any new warnings creep in?
:+1:

PR #1583, opened 2016-09-15
Fix: Premium label on Read next/latest + More ons

andygout:
cc @bjfletcher @andygnewman


andygnewman:
I see these are still using the articlePremium flag. May want to hold off merging until we know flag strategy for the test.

andygout:
Just spoke to @mmockett seconds ago and we're going to replace `articlePremiumLabel` with `premiumContentIndicator` because less associated to a single app and it needs changing in less places ;-)

PR #1586, opened 2016-09-16
fix: n-content-model consider datePublished to be lastPublished

matthew-andrews:
…… but there are loads of other bits ……


matthew-andrews:
😢

andygnewman:
Ha, I was also in the middle of doing that (plus was tidying up some of the code) thought it was a bizarre coincidence you asking about the date names, now it makes sense - will persevere with the tidy up.

PR #1593, opened 2016-09-19
add system code for the app

bjfletcher:
/cc @ifyio

/fyi @andygnewman @andygout ... a new requirement is to add system code to health checks, and, using n-express, this is how to do it


ifyio:
👍

PR #1597, opened 2016-09-20
Remove unnecessary babel transforms post node 6 upgrade

i-like-robots:
Also updates node-sass to a node-6 compatible version that won't spit out hundreds of lines of C internals!


i-like-robots:
CC @bjfletcher

PR #1598, opened 2016-09-20
add markets data link colour

leggsimon:
@i-like-robots

### DIFF
--- a/client/components/article/_main.scss
+++ b/client/components/article/_main.scss
@@ -190,4 +190,10 @@
 	}
 }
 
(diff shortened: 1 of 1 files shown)


client/components/article/_main.scss
```diff
@@ -190,4 +190,10 @@
 	}
 }
 
+// Markets Data Company Links
+[data-symbol] {
+	color: getColor('claret');
+	border-bottom-color: inherit;
+}
```
i-like-robots: hover/focus?

PR #1600, opened 2016-09-21
Add a description field for SEO love

richard-still-ft:


server/controllers/article.js
```diff
@@ -99,6 +99,14 @@ module.exports = function articleV3Controller (req, res, next, content) {
 	decorateMetadataHelper(content);
 	content.isSpecialReport = content.primaryTag && content.primaryTag.taxonomy === 'specialReports';
 
+	// Setup the description field
+	content.description = '';
+	if (content.summaries) {
```
matthew-andrews: i'd suggest swapping this logic around because `summaries` is deprecated

matthew-andrews:
it's good, suggested a small tweak

+1

server/controllers/article.js
richard-still-ft: right o

matthew-andrews:
💃

PR #1603, opened 2016-09-21
Bring a random tour tip into the article page

laurieboyes:
Does its very best to find a space, that doesn't cause weird blank space before an ad or quote or whatever. Makes room for light sign up iff the user is anonymous.

Behind both the `nextFtTour` and `nextFtTourTipArticlePage` flags

screen shot 2016-09-22 at 16 45 30


client/components/tour-tip/index.js
```diff
@@ -0,0 +1,12 @@
+export default {
+
+	init() {
+		const tipContainer = document.querySelector('.js-tour-tip-container');
+		const tip = document.querySelector('.tour-tip');
+
+		if(tip && tipContainer) {
+			tipContainer.appendChild(tip);
+			tipContainer.classList.remove('n-util-visually-hidden');
```
keirog: I reckon `n-util-hide` is more appropriate here

server/transforms/tour-tip.js
```diff
@@ -0,0 +1,24 @@
...
+		return $;
+	}
+
+	const pars = $('p');
+
+	positionComponent(1);
+
+	return $;
+
+	function positionComponent (position) {
```
keirog: my head = 💥

server/transforms/tour-tip.js
```diff
@@ -0,0 +1,24 @@
...
+		pars.each((index, par) => {
+			let indexMatches = ((index + 1) >= position);
+			let isOrphan = !par.parent;
+			let hasNextP = (par.next && par.next.name === 'p');
+			if (indexMatches && isOrphan && hasNextP) {
+				$(par).after('<div class="p402_hide article__body__tour-tip-container n-util-visually-hidden js-tour-tip-container"></div>');
```
keirog: `n-util-hide` here too, obvs

views/partials/tour-tip.html
```diff
@@ -0,0 +1,3 @@
+<div class="n-util-visually-hidden">
```
keirog: `n-util-hide` here too, obvs

phamann:
I'm confused as to why you created a container within the transform and then right out the component within the main template and use client-side JS

server/transforms/tour-tip.js
phamann: Why don't you just call `getRandomTourTipHtml()` here rather than creating a container?

views/content.html
```diff
@@ -54,6 +54,7 @@ <h2 class="sample-articles__ribbon__banner__heading" data-o-grid-colspan="12 M7
 						{{#if lightSignup.show}}
 							{{> light-signup tag=primaryTag articleUuid=id isInferred=lightSignup.isInferred}}
 						{{/if}}
+						{{> tour-tip tourTipHtml}}
```
phamann: This doesn't need to be here, as this will write out a tip for all articles regardless of whether the transform logic passes.

server/lib/get-random-tour-tip-html.js
```diff
@@ -0,0 +1,12 @@
+const cheerio = require('cheerio');
+const ReactServer = require('react-dom/server');
+const React = require('react');
```
keirog: Do we need to pull in React here? It doesn't seem to get referenced

laurieboyes: Yeah I was confused about that too but it falls over without it ¯\_(ツ)_/¯

laurieboyes:
@keirog thank you! I've had a go at making the meaning of significance a liiiittle bit clearer

keirog:
🚢

PR #1606, opened 2016-09-22
pa11y

lc512k:
basic a11y checks in CI. errors won't break the build


andygnewman:
Should those public files have been .gitignored?
Looks like package.json has some merge conflict leftovers in it.

PR #1614, opened 2016-09-27
Add click tracking to smartmatch

adgad:
@VladDubrovskis @andrewgeorgiou1981 @gvonkoss

### DIFF
--- a/client/components/ads/promoted-content.js
+++ b/client/components/ads/promoted-content.js
@@ -40,6 +40,9 @@ const getSmartmatchData = (adUnit, dfpResponse) => {
 		if(data && data.type && data.title) {
 			//add the DFP impression URL to smartmatch impression URLS
 			data.impressionURL = [].concat(data.impressionURL, dfpResponse.impressionURL);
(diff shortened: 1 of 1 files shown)


client/components/ads/promoted-content.js
```diff
@@ -79,10 +82,6 @@ const handleResponse = (el, response) => {
 		return;
 	}
 
-	if(response.url) {
```
VladDubrovskis: how come you removing this? as this was added in last PR:
https://github.com/Financial-Times/next-article/pull/1613/files#diff-e4571fff967851fcd44937f32f66f9d5R82

adgad: So I added it because the URL we get from Smartmatch is encoded, so had to de-encode it.

BUT then realised we also need DFP click tracking, which requires the URL to be encoded anyway

VladDubrovskis: OK, not sure I fully understand but as long as you are aware of this guess it good to go

PR #1616, opened 2016-09-27
chore(anon-sample-articles): remove expired code

keithamus:
This feature is no longer in use. Once this is merged I will also delete the feature flag.

Can @ironsidevsquincy review please?


keithamus:
The code in VCL was just opting users into next if they have a `segmentId`. Safe to keep for future features needing `segmentId`

PR #1623, opened 2016-10-03
suppress tips on listed articles

laurieboyes:


laurieboyes:
I've been talked out of this and I'm quite relieved

PR #1627, opened 2016-10-03
update the styles on article page for paid post and special reports

VladDubrovskis:
/cc @adgad 
screen shot 2016-10-04 at 12 41 55
screen shot 2016-10-04 at 12 42 58
screen shot 2016-10-04 at 12 43 07


client/components/ads/promoted-content.js
```diff
@@ -107,6 +107,7 @@ const handleResponse = (el, response) => {
 	};
 
 	ReactDOM.render(<components.Content {...props} />, el);
+	debugger;
```
adgad: eek!

client/components/ads/main.scss
```diff
@@ -60,6 +60,29 @@ $ad-medium-breakpoint: 760px;
 
 .promoted-content {
 	display: none;
+	border: 1px solid getColor('pink-tint3');
+	margin-top: 32px;
```
adgad: Could we just make this margin: 32px 0;

PR #1632, opened 2016-10-06
Run the tour-tip dismisser

keirog:
[Related n-ui PR](https://github.com/Financial-Times/n-ui/pull/543)


client/components/tour-tip/main.scss
```diff
@@ -6,11 +6,11 @@
...
 @include nUiCriticalEnd();
 
-
 .tour-tip--article-page {
 	@include nContentBoxInline();
-	margin-bottom: $spacing-unit;
 	border: 1px solid getColorFor(card, border);
-	display: block;
 	clear: left;
+	display: block;
```
debugwand: not familiar with `nContentBoxInline` but is sounds like it is the opposite of display block? or is it box-sizing?

keirog: I didn't write that bit, so not familiar either but it doesn’t look related: https://github.com/Financial-Times/n-content-body/blob/master/scss/_mixins.scss#L36

PR #1633, opened 2016-10-07
added barely minimum structured data

wheresrhys:
@mmockett this should help fix articles getting the wrong date in google. But there's lots more that could be added, which would be good for seo (not got the time or inclination to implement it all at present). https://developers.google.com/search/docs/data-types/articles http://schema.org/NewsArticle


views/partials/rel-canonical.html
```diff
@@ -1,3 +1,4 @@
 {{#if webUrl}}
 <link rel="canonical" href="{{canonicalUrl}}">
+<meta itemprop="mainEntityOfPage" content="{{canonicalUrl}}">
```
i-like-robots: don't know this one... the canonical above must take precedence though?

wheresrhys:
I think someone else has fixed this

PR #1634, opened 2016-10-10
[WIP] Remove xml fields

andygnewman:
cc: @i-like-robots 

Remove dependency on XML fields from ES.

Should wait for bodyHTML for V1 articles to be fully deployed, before merging in.


server/controllers/article.js
```diff
@@ -104,12 +104,9 @@ module.exports = function articleV3Controller (req, res, next, content) {
 		res.set('X-Robots-Tag', 'noindex');
 	}
 
-	// If no bodyHTML, revert to using bodyXML
-	const contentToTransform = content.bodyHTML || content.bodyXML;
-
 	// Apply content and article specific transforms to bodyHTML
-	if (contentToTransform) {
-		Object.assign(content, transformArticleBody(contentToTransform, res.locals.flags, {
+	if (content.bodyHTML) {
```
i-like-robots: Remove the conditional entirely?

test/server/controllers/article-helpers/myft-metadata.test.js
```diff
@@ -7,7 +7,6 @@ const fixtureEsFound = require('../../../fixtures/v3-elastic-article-found').doc
 
 const subject = proxyquire('../../../../server/controllers/article', {
 	'./article-helpers/suggested': () => Promise.resolve(),
-	'../transforms/article-xslt': (article) => Promise.resolve(article.bodyXML),
```
i-like-robots: oh wow

andygnewman: Was that a general 'oh wow' or one that should result in a further code change?

i-like-robots: Just amazed to see it here!

PR #1635, opened 2016-10-10
Remove n-image from babelified list of modules

andygnewman:
cc: @i-like-robots 

Now n-image pre-transpiled.

Bumped up minimum n-image version to pre-transpiled version.


i-like-robots:
For reasons I'm afraid I haven't been able to work out this fixed your build: https://github.com/Financial-Times/n-image/pull/43

PR #1642, opened 2016-10-13
Ads - fix bug with instream

andrewgeorgiou1981:
Ads; if the last element in the article body is not a paragraph then we no longer wish to insert an ad. This has been done to prevent layout issues if the last element is an absolute positioned


adgad:
👍 with minor comment

PR #1646, opened 2016-10-18
add the ssticky header style and put it as first thing after body ele…

VladDubrovskis:
…ment
/cc @adgad @andrewgeorgiou1981 @gvonkoss


andrewgeorgiou1981:
👍

PR #1647, opened 2016-10-19
Refactor lazy loading comments to allow for edge cases

tavvy:
Previously scenarios in which the user managed to "avoid" the
intersection observer (either scrolling too fast, or refreshing
from the bottom of the page) the comments would fail to load.

This change adds an additional check which happens on the very
first page scroll event (also fired when refreshed). If the check
finds the user is already below the intersection we would have
observed then we load the comments and dont init the observer.


tavvy:
The above PR isnt going to catch the scenario where the user follows the comments link (or any link on the page that will cause them to "jump" too far). Working on an additional fix now

tavvy:
@wheresrhys I'm aware you predominantly worked on this. Do you mind taking a look?

akinnane:
I've tested this on prod. It doesn't work as expected on firefox browsers. 

Browsing to todays(24/10) big read:
https://www.ft.com/content/5746e6dc-96b1-11e6-a1dc-bdf38d484582

Firefox OSX & WIndows 10 - 49.0.2
I find that the comments section loads in as soon as the page is loaded. There is no scroll action needed to initiate the lazy loading of livefyre comments. This happens at all page widths (Minimum width to full screen). I would expect the comments not to load until I had scrolled to within 600px of either the ".n-content-copyright" footer element or the "#comments" element. This is not the case. I have attached a screenshot.
screen shot 2016-10-24 at 10 27 13

Chrome OSX - 53.02785.143
Works as expected.

Safari OSX - 9.12
Works as expected.

Chrome - Andriod 53.0.2785.124
Comments can take a while to load but will load eventually (probably network related, much faster over wifi instead of 4G).

Chrome - Win10 50.2661.75
Works as expected.

Edge - Win10 - 25.10586.0.0
Works as expected

Opera - Win 10 - 40.0.2308.90
Works as expected.

PR #1648, opened 2016-10-24
Always use the same cachebuster for smartmatch

adgad:
@VladDubrovskis @andrewgeorgiou1981 @gvonkoss 

Based on their email.


VladDubrovskis:
does this need to go in any other apps?

adgad:
Nope, smartmatch currently only articles.

PR #1651, opened 2016-10-25
Prevent a dodgy tour tip positioning edge case

laurieboyes:
As well as the check for 3 following paragraphs, ensure the first child of each of these paragraphs is a text node. Prevents this sort of problem, where p tags are wrapped around non-p elements:

screen shot 2016-10-25 at 14 16 28

@i-like-robots had to fall back on digging around in the object again so, if you get chance, would be grateful for any cheerio wisdom you might have to impart. What I was looking for specifically was a way of widening the `:not(p)` selector to `:not(p) or (p with no text-node child)` but had no joy.


server/transforms/tour-tip.js
```diff
@@ -40,10 +40,14 @@ function positionComponent ($, position) {
 
 		const p = paragraphs.eq(i);
 
-		const hasThreeFollowingPs = p.nextUntil(':not(p)').length >= 3;
+		const followingPs = Array.from(p.nextUntil(':not(p)'));
+		const hasThreeFollowingPs = followingPs.length >= 3;
+		const followingThreePsHaveDirectText = followingPs.slice(0,3).reduce((soFar, followingP) => {
+			return soFar && followingP.children[0] && followingP.children[0].type === 'text';
+		}, true);
```
i-like-robots: There may be sugar for this...

i-like-robots: Maybe (no need to array.from, collections have a filter method already): 

```js
const hasThreeFollowingPs = p.nextUntil(':not(p)').filter((i, el) => $(el).children().length === 0)
```

laurieboyes: Gasp! Oh wait `.children()` doesn't return text nodes! Amazing, all my tests pass, you win.

PR #1652, opened 2016-10-26
FT Labs experiments with Spoken Layer integration

seanmtracey:
Markup + dependencies enabling playback of recorded version of article.


client/components/ftlabsSpokenLayer/main.scss
```diff
@@ -0,0 +1,3 @@
+#sl-player {
+	z-index: 1 !important;
```
andygnewman: don't use important, just use a good z-index

seanmtracey: The z-index is over-ridden by inline styles that the third party adds to the injected elements (they set a z-index of 999999999). Without the `!important` tag, the inline styles over ride the CSS styles and the widget appears over the top of the navigation bar as the user scrolls down.

client/main.js
```diff
@@ -72,6 +73,11 @@ bootstrap(nUiConfig, ({flags, mainCss}) => {
 		if (flags.get('articleComments') && document.querySelector('#comments')) {
 			commentsInit();
 		}
+
+		if(flags.get('ftlabsSpokenLayer')){
```
andygnewman: I could be really picky and say that there should be a space between the if & the (

PR #1654, opened 2016-10-28
added csp header

gvonkoss:
@VladDubrovskis :D


server/controllers/article.js
```diff
@@ -72,6 +72,8 @@ module.exports = function articleV3Controller (req, res, next, content) {
 	res.vary('ft-barrier-type');
 	res.vary('x-ft-auth-gate-result');
 
+	res.vary('ft-allow-csp');
```
VladDubrovskis: Don't vary on it, just use `.set` instead.

Also worth adding a test to make sure header is actually set

PR #1655, opened 2016-10-31
Update version for comment count and supporting

debugwand:
Despite having a caret version, o-comment-count was not updating to 0.3.x - see https://circleci.com/gh/Financial-Times/next-article/4918

This ensures that the right version is called in and makes sure the related o-comment-x dependencies are also up to date


GlynnPhillips:
@debugwand Obviously there's nothing wrong with this comment but are we also looking at why this didn't behave as expected?

debugwand:
probably building with cache

PR #1656, opened 2016-11-01
Added property to content object.

seanmtracey:
We don't want to load the Spoken Layer script to check whether or not there is an audio file available for an article on every single page load. This PR adds a check for a metadata tag ('Audio articles') and only adds the `ftlabsSpokenLayer` partial if the content has been picked for recording by FirstFT.

Also, removed the FT Labs gift article header while I was here.


server/controllers/article.js
```diff
@@ -176,6 +171,10 @@ module.exports = function articleV3Controller (req, res, next, content) {
 		isInferred: res.locals.flags.lsuInferredTopic
 	};
 
+	content.isAudioArticle = content.metadata.filter(foo => {
+		return foo.name === 'Audio articles';
+	}).length > 0;
+
```
andygnewman: OK, lets refactor that;

```
content.isAudioArticle = content.metadata.some(tag => tag.prefLabel === 'Audio articles');
```
`some` returns a boolean based on whether at least one matches the condition.
`prefLabel` is the name field for the tags in the metadata, but do you mean this or `taxonomy` - how is this metadata being set?
also should wrap this in your flag condition - whilst not a massive issue, I think that checking the flag will be quicker than doing the `some` before it gets more widely rolled out.

views/content.html
```diff
@@ -37,9 +37,11 @@
...
 					</div>
 
-					{{#if @root.flags.ftlabsSpokenLayer}}
-						{{> ftlabsSpokenLayerPlayer}}
-					{{/if}}
+					{{#ifEquals isAudioArticle true}}
+						{{#if @root.flags.ftlabsSpokenLayer}}
+							{{> ftlabsSpokenLayerPlayer}}
+						{{/if}}
+					{{/ifEquals}}
```
andygnewman: We can refactor this too;
```
{{#ifAll isAudioArticle @root.flags.ftlabsSpokenLayer}}
    {{> ftlabsSpokenLayerPlayer}}
{{/ifAll}}
```
`ifAll` helper https://github.com/Financial-Times/n-handlebars#ifall

server/controllers/article.js
seanmtracey: We mean `prefLabel`. We didn't realise name didn't always exist across the different content API requests. The metadata is being applied to the article in Methode.

views/content.html
seanmtracey: 👍

PR #1658, opened 2016-11-03
Teaserified article

adgad:
@keirog @bjfletcher 
Behind the nTeaserArticle flag.

TODO:
- [ ] Paid posts
- [ ] Possible reskin of non-article inline components


client/components/ads/promoted-content.js
```diff
@@ -111,8 +116,16 @@ const handleResponse = (el, response) => {
 		standfirst: { show: { default: true } }
 	};
 
-	ReactDOM.render(<components.Content {...props} />, el);
+
+	if(flags.nTeaserArticle) {
+		//el.innerHTML = teaserTemplate(response);
+		ReactDOM.render(<components.Content {...propsForReact} />, el);
+	} else {
+		ReactDOM.render(<components.Content {...propsForReact} />, el);
```
bjfletcher: both if and else the same?

adgad: Yup, temporarily commented this out until we get the client-side n-teaser stuff working

adgad: I'll roll it back for now so it's less confusing

client/main.scss
```diff
@@ -135,32 +140,62 @@ $o-video-is-silent: false;
...
-			width: 100pt;
-		}
+.o-teaser .n-image-wrapper--placeholder {
+	height: 0;
+}
 
-		.n-content-tags__product-tour {
-			clear: both;
-		}
+//make the myFT follow box look okay with collection styles
```
bjfletcher: all these styling, I was hoping would be in n-teaser/n-teaser-collection, rather than in the individual apps like in here...

adgad: Yes I think so. At the moment this is all overriding styles that the non-flag version has. I'm hopeful that once the teaser stuff is default these styles can be tidied up.

test/server/transforms/related-article-to-teaser.test.js
```diff
@@ -0,0 +1,116 @@
...
+				'<p>Graham Tatomer worked at Austrian Riesling producer Emmerich Knoll and now fashions this example from the old vines of the Kick-on Ranch in Santa Barbara (£26.95, Roberson)</p>' +
+				'</div>' +
+				'</aside>',
+				{ decodeEntities: false }
+			);
+			relatedArticleToTeaserTransform($, { nTeaserArticle: true });
+			teaserStub.callCount.should.equal(0);
+		});
+	});
+});
```
bjfletcher: is @andygnewman aware of this article to teaser transform, or maybe he wrote this. :)

adgad: Aware but would be good to get his eyes :)

server/controllers/related/more-on.js
```diff
@@ -75,15 +75,20 @@ module.exports = function (req, res, next) {
 			moreOnArticlesArray[moreOnIndex] = dedupe(moreOnArticlesArray[moreOnIndex])
 				.map(article => Object.assign(article, contentDecorator(article)));
 
+			if(res.locals.flags.nTeaserArticle) {
+
+				return res.render('partials/teaser-collections/1-3', { items: moreOnArticlesArray[moreOnIndex] });
```
bjfletcher: I understand. The same thinking was had when frontpage was made... it was then very difficult for stream to reuse things from frontpage. I think it'd be really nice to get UX to come up with "off the shelf" components in n-teaser-collection, that UX people can easily ask us to use in different pages/apps. I'd much rather that than having many different, customised collections.

adgad: Agreed. The good news is there's only half a dozen so far and they're far simpler than n-section :)

PR #1679, opened 2016-11-16
Acquire: More-ons data via next-api query

andygout:
cc @ironsidevsquincy 

- [x] Factor in existing usage of: `// Get +1 for de-duping parent article` `count: count + 1`.
- [x] Fix tests: https://github.com/Financial-Times/next-article/blob/master/test/server/controllers/related/more-on.test.js.
- [x] NEXT_API_KEY added in this PR: https://github.com/Financial-Times/next-config-vars/pull/488


server/lib/fetch-graphql-data.js
```diff
@@ -0,0 +1,13 @@
+const fetchres = require('fetchres');
+
+module.exports = (query, variables = {}) => {
+	return fetch(`https://next-api.ft.com/v1/query?query=${query}&variables=${JSON.stringify(variables)}&source=next-article`, {
+			method: 'GET',
+			headers: {
+				'X-Api-Key': process.env.NEXT_API_KEY
```
andygnewman: Article doesn't yet have a NEXT_API_KEY so will need to be set-up in config vars

server/controllers/related/more-on.js
```diff
@@ -1,28 +1,69 @@
...
 const logger = require('@financial-times/n-logger').default;
+const fetchGraphQlData = require('../../lib/fetch-graphql-data');
 const NoRelatedResultsException = require('../../lib/no-related-results-exception');
 const contentDecorator = require('@financial-times/n-content-decorator');
 const ReactServer = require('react-dom/server');
 const React = require('react');
 const getSection = require('../../../config/sections');
 const Section = require('@financial-times/n-section').Section;
 
+const moreOnsQuery = () => (`
```
andygout: Yup - will change name to `relatedContentQuery` will seems to be a catch-all term for the instances in which we might require it?

server/lib/fetch-graphql-data.js
andygout: Remembered that we have not yet enforced the presence of NEXT_API_KEY so for the purpose of More-Ons it's unnecessary, but am guessing we'll want to use this same function for getting article data in which instance we'll need that enforcement, so will add to `next-config-vars`.

andygnewman:
Nice. Will look to refactoring that graph ql query for the new content model when I update the n-teaser version. 👍

server/controllers/related/more-on.js
```diff
@@ -1,28 +1,30 @@
...
+			return articles.search
 				.filter(article => article.id !== parentId)
 				.slice(0, count)
-		});
+		})
+		.catch(err => {
+			logger.error(err);
+			return {
+				articles: []
+			}
```
ironsidevsquincy: shouldn't this just be an array?

server/controllers/related/more-on.js
```diff
@@ -1,28 +1,30 @@
...
 
-
 const getArticles = (tagId, count, parentId) => {
-	return api.search({
-		filter: [ 'metadata.idV1', tagId ],
-		// Get +1 for de-duping parent article
-		count: count + 1
-	})
+	return fetchGraphQlData(relatedContentQuery, { tagId, limit: count + 1 })
 		.then(articles => {
```
ironsidevsquincy: think this is badly named now. using destructuring,

```.then(({ search: articles = []} = {}) => {```

means you wont need to change anything below

PR #1685, opened 2016-11-18
Amend: get client side story package from next-api

andygnewman:
cc: @ironsidevsquincy @andygout 

This swaps from next-ft-api-client to next-api for client side article story package.


server/controllers/related/story-package.js
```diff
@@ -7,26 +6,24 @@ const ReactServer = require('react-dom/server');
...
-		.then(function (articles) {
-			if (!articles.length) {
+	return fetchGraphQlData(storyPackageQuery, { uuid: req.params.id, limit: count })
+		.then(({article}) => {
+			if (!article.storyPackage || !article.storyPackage.length) {
 				throw new NoRelatedResultsException();
 			}
-			if (articles.length % 2 === 0) articles.length --;
-			return articles.map(article => Object.assign(article, contentDecorator(article)));
+			return article.storyPackage.map(article => Object.assign(article, contentDecorator(article)));
```
ironsidevsquincy: even though it's fine because it's scoped, I try to avoid using the same variable name for things when the scope is that close (just looks a bit confusing all those `article`s

could go destruct the `then` argument

```.then({ article: { storyPackage = [] } = {}) => {```

No need for the outer `article` then (or the `!article.storyPackage` check)

andygnewman: I just thought about that and in my head I said - "I bet Darren picks up on that"

ironsidevsquincy: I'm becoming predictable...

Issue #1680, opened 2016-11-16
Use `next-api` to acquire content


andygout:
Article: https://github.com/Financial-Times/next-article/pull/1673
More-Ons: https://github.com/Financial-Times/next-article/pull/1679
Story Package: TO DO
Read Next: TO DO

PR #1691, opened 2016-11-21
Amend: migrate suggested reads to next api

andygnewman:
cc: @ironsidevsquincy @andygout 
Final part of the migration (https://github.com/Financial-Times/next-article/issues/1680) of related content elements of the article page to next-api (graphql)


server/controllers/article-helpers/onward-journey.js
```diff
@@ -0,0 +1,33 @@
...
+
+			return {
+				readNext: getReadNext(topicArticles, packageArticles, publishedDate),
+				suggestedReads: getSuggestedReads(topicArticles, packageArticles)
+			};
+		})
+		.catch(
+			error => logger.warn('Fetching onward journey data failed.', error.toString())
+		);
+};
```
ironsidevsquincy: will probably get a clash with the PR I just merged - https://github.com/Financial-Times/next-article/blob/e485e394e7a69b790ea039452939f7acff7f2e28/server/controllers/article-helpers/read-next.js

Update what was read-next to handle articles without a `primaryTag` (like video content is currently)

server/controllers/podcast.js
```diff
@@ -41,18 +40,13 @@ module.exports = function podcastLegacyController (req, res, next, payload) {
...
-
-		asyncWorkToDo.push(
-			readNextHelper(payload.id, storyPackageIds, payload.primaryTag, payload.publishedDate).then(
-				article => payload.readNextArticle = article
-			)
+			getOnwardJourneyArticles(payload.id, payload.publishedDate)
+				.then(onwardJourney => {
+					payload.readNextArticle = onwardJourney && onwardJourney.readNext;
+					payload.readNextArticles = onwardJourney && onwardJourney.suggestedReads;
+				})
```
ironsidevsquincy: will the new `video` controller also need updating?

andygnewman: Was going to suggest that when you write the video controller you may want to refactor the three controllers into a single process (given the amount of duplication) 😄

PR #1710, opened 2016-12-02
Video player pass the second

rowanbeentje:
Bit of a whopper - too much figuring out of Next as I went along to break into smaller, more sensible commits, I'm afraid (which is also a shame in case the design later veers back to one of the two previous versions!).

The basic functionality is all here and working, but it's not fully complete: another pass of minor style fixes is needed once I've worked through the list, and a new version of o-video is needed before all the autoplay behaviour is there, but thought it best to get something in for now.


client/components/video/main.scss
```diff
@@ -0,0 +1,179 @@
+$video-background-primary: #252830;
+$video-grey-text: #999999;
```
ironsidevsquincy: can use o-color's grey-tint2 for this, e.g. `getColor('grey-tint2')`

server/controllers/related/video-up-next.js
```diff
@@ -0,0 +1,47 @@
+const logger = require('@financial-times/n-logger').default;
+const fetchGraphQlData = require('../../lib/fetch-graphql-data');
+const upNextContentQuery = require('../../graphql-queries/video-up-next');
+
+module.exports = function (req, res, next) {
+	res.unvaryAll('wrapper');
```
ironsidevsquincy: what does this do?

views/partials/body-video.html
```diff
@@ -1,14 +1,63 @@
...
+				<p class="video__byline">
+					{{byline}}
+				</p>
+
+				{{#if readNextArticles}}
+					<h4 class="video__related-articles-title">Related Articles</h4>
+					<ul class="video__related-articles-list">
+					{{#each readNextArticles}}
+						<li class="related_content__list-item" data-trackable="related-content | {{@index}}">
+							<a href="{{url}}">
```
ironsidevsquincy: add a `data-trackable` attribute to this, so it can be tracked

server/controllers/related/video-up-next.js
rowanbeentje: This seems to get used on all the fragment responses, and I assume allows cache targetting for things that don't include wrapper html...?  Not sure though :)

rowanbeentje:
Not sure if you're done reviewing, but as you've approved I'll hit merge; many more changes to come and I'll include further cleanups as I see comments :)

PR #1714, opened 2016-12-05
Video player page autoplay and rollovers

rowanbeentje:
Applies a number of design tweaks from Rebecca, fixes n-teaser compatibility, and enables the placeholder/autoplay/play-next functionality.  The latter needs some additional hooks in o-video in the long term to be less hacky...

### DIFF
--- a/client/components/video/autoplay.js
+++ b/client/components/video/autoplay.js
@@ -1,8 +1,73 @@
 'use strict';
 
-module.exports = () => {
-	const primaryVideoPlayer = document.querySelector('.content__video [data-o-component="o-video"]');
-	if (!primaryVideoPlayer) {
+const lazyLoadImages = require('n-image').lazyLoad;
+
+const videoPlaceholderElement = document.querySelector('.video__placeholder');
+let upNextTimer = false;
+
+module.exports = oVideo => {
+	oVideo.init().then(() => {
+		oVideo.play();
+		videoPlaceholderElement.addEventListener('click', oVideo.play.bind(oVideo));
+	})
+};
+
+module.exports.init = () => {
+	document.body.addEventListener('oTracking.event', eventForwarder);
+};
+
+function eventForwarder (event) {
+	const trackingDetails = event.detail;
+	if (!trackingDetails || trackingDetails.category !== 'video') {
 		return;
 	}
-};
+
+	switch (trackingDetails.action) {
+		case 'adStart':
+		case 'playing':
+			videoPlaying();
+			break;
+
+		case 'pause':
+			videoPaused();
+			break;
+
+		case 'ended':
+			videoEnded();
+			break;
+
+		default:
+			// Ignore event
+	}
+}
+
+function videoPlaying () {
+	videoPlaceholderElement.classList.add('video__placeholder__played');
+	videoPlaceholderElement.classList.add('video__placeholder__playing');
+	clearTimeout(upNextTimer);
+}
+
+function videoPaused () {
+	videoPlaceholderElement.classList.remove('video__placeholder__playing');
+}
+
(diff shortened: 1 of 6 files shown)


client/components/video/autoplay.js
```diff
@@ -1,8 +1,73 @@
...
+		const rolloverUpNext = upNextItem.cloneNode(true);
+		rolloverUpNext.removeAttribute('data-o-grid-colspan');
+
+		upNextPlaceholderSlot.innerHTML = '';
+		upNextPlaceholderSlot.appendChild(rolloverUpNext);
+
+		videoPlaceholderElement.classList.add('video__placeholder__ended');
+		videoPlaceholderElement.classList.remove('video__placeholder__playing');
+
+		lazyLoadImages();
```
rowanbeentje: This is the bit that doesn't seem to be working reliably, for some reason - could it be because the image is still in the process of fading in at this point?

client/components/video/autoplay.js
```diff
@@ -1,8 +1,73 @@
...
+};
+
+module.exports.init = () => {
+	document.body.addEventListener('oTracking.event', eventForwarder);
+};
+
+function eventForwarder (event) {
+	const trackingDetails = event.detail;
+	if (!trackingDetails || trackingDetails.category !== 'video') {
 		return;
```
ironsidevsquincy: will this turn off all non-video tracking on the page?

client/components/video/autoplay.js
```diff
@@ -1,8 +1,73 @@
...
+			videoEnded();
+			break;
+
+		default:
+			// Ignore event
+	}
+}
+
+function videoPlaying () {
+	videoPlaceholderElement.classList.add('video__placeholder__played');
```
ironsidevsquincy: what does `played` signify?

client/components/video/autoplay.js
ironsidevsquincy: yeah, possibly the item is still being added to the page. Maybe try adding a hacky setTimeout. Can also just scope the lazy loading to this element (instead of the whole dom)

```lazyLoadImages({ root: upNextPlaceholderSlot });```

File: client/components/video/main.scss [hunk 4]
```diff
@@ -33,11 +34,110 @@ $video-grey-text: oColorsGetPaletteColor('grey-tint2');
...
+	visibility: visible;
+
+	> .o-video__info {
+		padding: 25px 20px;
+	}
+
+	&.video__placeholder__played {
+		bottom: 40px;
+		touch-action: none;
+		background: linear-gradient($placeholder-background-color, $placeholder-background-color, $placeholder-background-color, transparent); // Fades out the end
```
ironsidevsquincy: guessing this is just an extra bit oomph, as in browsers that don't understand `linear-gradient` are still usable

client/components/video/autoplay.js
rowanbeentje: This is just an additional listener for tracking events as a workaround for there not being any other video events available right now (will add a list in o-video at which point this will all become cleaner).  We only really care about video events for the purposes of adding more hooks :)

client/components/video/main.scss
rowanbeentje: This is all a hack so that the native controls of the video (at the bottom in all browsers, I sincerely hope) are still usable when the overlay is showing.  This will properly be fixed when o-video gets FT-designed HTML controls, but I don't see this happening in the near future...

client/components/video/autoplay.js
rowanbeentje: Hasn't fixed it reliably :(  Worth leaving in the scoping though; I'll revisit this later.

PR #1717, opened 2016-12-05
Video player page visual improvements

rowanbeentje:


client/components/video/main.scss
```diff
@@ -70,7 +74,7 @@ $placeholder-background-color: rgba(oColorsGetPaletteColor('claret'), 0.65);
 
 	&.video__placeholder__played {
 		bottom: 40px;
-		touch-action: none;
+		pointer-events: none;
```
rowanbeentje: I was using entirely the wrong rule for what I was intending to do, oops.

PR #1723, opened 2016-12-06
Affinity component

pixelandpage:
Injects the affinity recommendations into the right-hand rail and the bottom sections of article pages using n-teaser templates.


views/partials/right-hand/right-hand.html
```diff
@@ -7,4 +7,12 @@
 		{{> right-hand/suggested-reads-teaser-collection items=readNextArticles topic=readNextTopic }}
 	{{/if}}
 
+	{{#if @root.flags.affinity}}
+		<div class="o-teaser-collection" data-trackable="affinity">
+			<div id="affinity-rhs">[loading indicator to go here]</div>
```
paddyo: Maybe put the indicator there or delete this text.

PR #1750, opened 2017-01-05
Copy warning/char cap

aintgoin2goa:
@lc512k 
@tavvy 

Adding a message at the top if someone tries to copy/paste 30 words or more.
Works in all the browsers I've tested it in, but could use some thorough QA...


commuterjoy:
Maybe add it to the bottom?  It's an annoying thing.

aintgoin2goa:
It was a requirement to go at the top I think.  Based on this from Trello card. Also, if it goes at the bottom the user might not see it and email it off to someone.  Less annoying but more embarrassing?

upload_10_27_2016_at_3_19_40_pm

PR #1751, opened 2017-01-06
removed legacy design

gvonkoss:
this wasn't being used by the looks of things (it's been about a year and a half now), and considering RJ, there may be some design changes!

@andygnewman


andygnewman:
👍  once linted - feels nice to delete code...

PR #1758, opened 2017-01-10
Enhanced stylings for Spoken Layer Player

seanmtracey:
No urgency for merging.


andygnewman:
Sean,

There's a lot of condensing / nesting that you could do with the styling which would help with understanding what's happening here.

In addition the naming should be following BEM guidelines - http://getbem.com/introduction/

andygnewman:
Plus fix the linting

PR #1769, opened 2017-01-16
Make screen-reader experience less confusing

keirog:
Hides the following from screen readers:
- pull-quotes
- mid-content related-article boxes
- mid-content light signup form

This is to avoid breaks in the flow of the main content. Reading via screen reader is much less confusing when these breaks in content are removed.

Not an ideal solution as we’re hiding things that add richness and context, but seems the least bad option without redesigning the page.

Addresses https://github.com/Financial-Times/next-a11y/issues/12

@lc512k


server/transforms/related-article-to-teaser.js
```diff
@@ -39,6 +39,7 @@ module.exports = function ($) {
 		});
 
 		$el.addClass('n-content-related-box--no-border');
+		$el.attr('aria-hidden', 'true');
```
lc512k: This is the only one that worries me slightly, as we're simplifying the SR experience rather than fixing it. Ideally, these would be at the end of the article (after all the `<p>`s) so users can get to them, without interrupting them.

For the time being I'm happy to go with this. We might need a gh issue to revisit after accreditation

server/transforms/body.js
```diff
@@ -26,6 +27,7 @@ module.exports = function (body, flags, options) {
 	transform($bodyHTML, flags, options)
 		.with(relatedArticleToTeaser)
 		.with(relatedBoxExpander)
+		.with(pullquoteAriaHidden)
```
lc512k: ❤️

PR #1777, opened 2017-01-18
Rich content

TheoLeanse:
@adgad @gvonkoss @kavanagh 

Pushes experimental content that lives in capi test (or prod, because prod stuff gets published to test, too?) thru the next article app in two ways:

1) run the app as normal with environment variable `EXPERIMENTAL_CONTENT` (must be unset afterwards)
2) hit the app's new`/preview/<TEST_CONTENT_UUID>` endpoint


adgad:
🚢

adgad:
Also perhaps add a note to the README

PR #1785, opened 2017-01-23
Extras

gvonkoss:
@TheoLeanse  @adgad


client/components/article/topper/_main.scss
```diff
@@ -104,3 +107,36 @@
 		margin: 0;
 	}
 }
+
+@include oGridRespondTo($until: S) {
```
adgad: You could maybe just use colspans here - https://github.com/Financial-Times/o-grid#hiding-elements

PR #1788, opened 2017-01-24
removed much legacy

gvonkoss:
@andygnewman @adgad 
seeing as I'm deleting stuff both of you wrote, anyone care to weigh in? :D


adgad:
Can you also delete the server-side bit that sets `tagToFollow`?

andygnewman:
Presume you've checked with myFT about removing the tag to follow, that was something they built.

Should also get rid of the flag for articleSuggestedRead (not sure if it's used in the server side code to determine whether to run the code for it), but if it's the only show in town, probably no longer required.

Also is there any CSS you can delete?

commuterjoy:
"removed much legacy" sounds like a Trump slogan

gvonkoss:
@andygnewman as far as I can tell, a good bit of the suggestedRead includes read next and such, so best keep that flag for now?

@adgad i've removed all of the server-side code, and what was cropping up in n-concept, as well.

andygnewman:
@gvonkoss the comment about flag removal was more that if read next etc is the only option why have it controlled by flag vs just doing it anyway. The flag originally was for an AB test.

adgad:
LGTM

PR #1800, opened 2017-01-27
first working prototype of inline barriers on article page

tavvy:
How this would work
----

**ANON** article requests with **both** the header `ft-access-preview` and the flag `inArticlePreview:on` will be served a cut down version of the article with an inline barrier served at the end.

While we test, this is the only current use case and scenario. For now this is a proof of concept.

**Related PRs**
- https://github.com/Financial-Times/next-router/pull/58
- https://github.com/Financial-Times/ft.com-cdn/pull/306
- https://github.com/Financial-Times/next-preflight/pull/155

----

After testing we would remove the `inArticlePreview` flag requirement and adopt the following approach:

 - **ANON** article requests with `ft-access-preview` header will be served a cut down version of the article. That is all that header is responsible for, it allows for a separation of concerns. 

 - This allows the format of the inline barrier to be flexible (perhaps light signup, or "buy now" etc).
The type of barrier would be dictated by a combination of preflight cohort grouping (headers) and flags. 

The current (proof of concept) implementation uses a hardcoded `standard` setting for now.

----

**Notes:**
 - Article trimming is pretty simplistic for now
 - Hopefully this is fairly easily extensible
 - The inline barrier will **disable** the light signup component
 - Tests are required


client/components/inline-barrier/main.js
```diff
@@ -0,0 +1,31 @@
...
+const populateWithBarrier = (el) => {
+	return fetch('/products?fragment=true&inline=true&narrow=true', { credentials: 'same-origin' })
+		.then(response => {
+			if (response.ok) {
+				return response.text().then(html => {
+					el.insertAdjacentHTML('beforeend', html);
+					el.classList.remove('n-util-visually-hidden');
+				});
+			} else {
+				return response.text().then(() => {throw('Could not fetch barrier content')});
```
i-like-robots: no need to resolve the text if you're not using it

server/transforms/teaser-content.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+const asObj = ($) => ({ bodyHTML: $.html() });
+
+module.exports = function (bodyHTML, flags) {
+	const $ = cheerio.load(bodyHTML);
+
+	if (!flags.inArticleTeaser) return asObj($); // [1]
+
+	const $contents = $('*');
```
i-like-robots: I don't think this is necessary, you could just `const $content = cheerio.load(bodyHTML)`

views/partials/inline-barrier.html
```diff
@@ -0,0 +1 @@
+<div id="inline-barrier" class="n-util-visually-hidden" data-n-component="inline-barrier" data-n-type="{{type}}" data-trackable="inline-barrier"></div>
```
wheresrhys: what about core experience? Should there just be a simple paragraph explaining the article is curtailed because they don't have permission?

server/transforms/teaser-content.js
i-like-robots: Oh I see you're getting an array of _everything_... easier to just get top-level stuff?

tavvy: The bodyHTML returned by es has no containing div. Like you say this was a way of getting an array of the content.

andygnewman:
Not sure if you've checked out the user journey from the FastFT stream, whereby the article HTML is rendered within the page. 

With current barriers, it would just render the full barrier as this would have intercepted the article request.

So to make it work in this scenario, the inline barrier html should be returned within the fragment view as well as content view.

CSS should be OK as streams has n-barriers (though may need bumping)

tavvy:
Thanks all.

@andygnewman good point, I hadn't properly considered that view.
Currently the logic wont inject the barrier / cut the content if it is a fragment request. However this may poke a hole in the paywall (via the fastFT stream). I will check back in with the team.

keirog:
@lc512k @tavvy Is it too late for us to change the name? A teaser is already a thing 😬 🍰 
How do you feel about preview instead?

PR #1803, opened 2017-01-31
feat: redirect to placeholder url

TheoLeanse:
@adgad @i-like-robots 

depends on adding a type field in ES


adgad:
👍

adgad:
Also perhaps a test :)

PR #1809, opened 2017-02-01
fix error log

gvonkoss:
@adgad

### DIFF
--- a/server/lib/capi-image.js
+++ b/server/lib/capi-image.js
@@ -14,15 +14,9 @@ const fetchCapiJson = (endpoint) => {
 			}
 			return response.text()
 				.then((text) => {
-					const requestId = response.headers.get('x-request-id') || 'UNKNOWN';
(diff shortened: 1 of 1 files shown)


server/lib/capi-image.js
```diff
@@ -47,8 +41,8 @@ module.exports = function (image) {
 		.catch((err) => {
 			logger.error({
 				event: 'TOPPER_IMAGE_FETCH_FAIL',
-				error: err.toString(),
-				uuid: image
+				error: err.message.toString(),
+				uuid: image.id
```
ironsidevsquincy: n-logger 'understands' error objects if you pass them through as their own argument, e.g.

```logger.error({ event: 'TOPPER_IMAGE_FETCH_FAIL', uuid: image.id }, err);```

adgad: Fixed in a series of reckless commits to master.

ironsidevsquincy: the only way to code...

PR #1812, opened 2017-02-01
Use content type field

TheoLeanse:
Something to merge when the nice new `type` field is on all bits of content in ES

@adgad @i-like-robots


kavanagh:
:shipit:

i-like-robots:
http://next-elastic.ft.com/content/item/_search?q=_missing_:type 🍾

PR #1814, opened 2017-02-02
Get the Sepcial Report's label

ironsidevsquincy:
Make an extra request to next-api, to be safe


andygnewman:
Far better way of doing it. Will need to fix the test though.

PR #1819, opened 2017-02-02
Add closed-captions to video player

commuterjoy:
- [x] Fix cross-origin errors in `next-video-player`
- [x] Relies on release of https://github.com/Financial-Times/o-video/pull/87


server/controllers/video.js
```diff
@@ -33,6 +34,10 @@ module.exports = function (req, res, next, payload) {
 		};
 	}
 
+	if (res.locals.flags.closedCaptions) {
+		payload.captions = payload.attachments.filter(attachment => attachment.mediaType === 'text/vtt').reduce((a, b) => b, undefined);
```
ironsidevsquincy: does this reduce just get the last element?

commuterjoy: yes, will need to rework if we add multiple caption/subtitle tracks.

PR #1822, opened 2017-02-06
Add FinancierDisplayBold to topper

kavanagh:


adgad:
Hmm...isn't this going to cause FOIT on slower connections (well, all connections)?

kavanagh:
@adgad at least this works. probably going to do some more refactoring in another branch.

PR #1829, opened 2017-02-07
[WIP] Update to ElasticSearch for topper

adgad:
awaiting URL to be available in prod


server/controllers/article-helpers/topper-theme-map.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = (topper) => {
+	if(topper && topper.theme && themeImageRatio[topper.theme]) {
```
gvonkoss: is the themeImageRation living somewhere else?

adgad: Yup, just added it. I changed the check to themeImageRatio.hasOwnProperty(topper.theme), because `full-bleed-text` has themeImageRatio null, so (I think) that would never have worked?

PR #1830, opened 2017-02-08
Use alternative standfirsts, not bodyHTML

commuterjoy:
ref: https://github.com/Financial-Times/next-es-interface/pull/721/files


views/partials/body-video.html
```diff
@@ -57,7 +57,7 @@ <h1 class="video__placeholder__headline article-headline">
 				{{> header/content-video-header}}
 
 				<div class="video__description">
-					{{{bodyHTML}}}
+					{{alternativeStandfirsts.promotionalStandfirst}}
```
i-like-robots: If this is just text does it need to be in a paragraph?

PR #1831, opened 2017-02-08
Use Preact

bjfletcher:


andygnewman:
The only issue I can detect in using this in article is that the autofocus doesn't seem to be working. ie. when you add an extra email address input, the focus does not move to the last input as it does in the react environment.

I think @i-like-robots may be able to help suggest how to handle that in preact.

@bjfletcher were there other issues that you were concerned about in implementing this?

andygnewman:
cc: @adgad 

This PR migrates the email share functionality from React to Preact.
Plus brings in (behind a flag) the ability for users to send a custom message with the email (https://github.com/Financial-Times/n-email-article/pull/34)

bjfletcher:
🎉  Nice work Andy

i-like-robots:
\0/

PR #1835, opened 2017-02-09
Offset theme fixes

gvonkoss:
@kavanagh @adgad @TheoLeanse

### DIFF
--- a/client/components/article/topper/_main.scss
+++ b/client/components/article/topper/_main.scss
@@ -93,19 +93,12 @@
 .article-topper__extras {
 	display: flex;
 	padding-bottom: 5px;
+	width: 125%;
 
 	@include oGridRespondTo($until: S) {
 		flex-direction: column;
 		justify-content: center;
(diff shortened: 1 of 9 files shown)


client/components/article/topper/_main.scss
```diff
@@ -93,19 +93,12 @@
 .article-topper__extras {
 	display: flex;
 	padding-bottom: 5px;
+	width: 125%;
```
adgad: Magic number alert - why is it 125%?

client/components/article/topper/_mixins.scss
```diff
@@ -158,3 +158,16 @@ $colors: 'white', 'black', 'claret', 'blue', 'pink', 'slate';
 		z-index: 1;
 	}
 }
+
+@mixin tagBehaviourSplit {
```
adgad: Could we rename this to what the actual behaviour is, rather than where it is to be used (I don't really know what it is)?

PR #1838, opened 2017-02-09
handle an article in a series (behind flag)

TheoLeanse:
@kavanagh @adgad 

* Initial content package article navigation (behind flag)
* Initial content-package controller/view (behind flag)
* Moves onward-journey graphQL query from article/video/podcast controllers to negotiation
* (if flag on) adds contaiins/contains to the onward journey query


package-hack-list.js
```diff
@@ -0,0 +1,123 @@
+const books = {
```
adgad: Can probably delete this?

TheoLeanse: yes, will do

kavanagh:
ok, do this!

PR #1839, opened 2017-02-09
legal copy notice fix for Text To Speech (TTS)

lc512k:
Expands on https://github.com/Financial-Times/next-article/pull/1836 and addresses copy via `contextmenu`

@keirog the diff isn't pretty, best to look at https://github.com/Financial-Times/next-article/blob/ff0b631c559d2c53260bd0634db7a0db2ffdd377/client/components/legal-copy/main.js#L66-L115 for context

tl;dr `alt+esc` is the shortcut for text-to-speech on the mac, but it also triggers a synthetic `ctrl+c` and effectively the `copy` event on chrome 56 and below. Previous functionality meant that users were read out the copyright warning instead of the article. 

The previous PR added a workaround to add the copyright warning for `ctrl-c` only when not triggered by `alt`. This PR extends the functionality to `copy` called from the context menu.

More verbose than the original `document.body.addEventListener('copy', onCopy);` and still allows copying via the edit menu (though you'd have to _guess_ that) but it means TTS users can still read articles.  

We can (mostly) have our 🍰 and eat it too now.


client/components/legal-copy/main.js
```diff
@@ -60,26 +61,55 @@ module.exports = function (flags) {
...
+function interceptKeys (e) {
 
-	if (evt.keyCode === ALT) {
+	// Remember if alt was pressed
+	if (e.keyCode === ALT) {
 		altPressed = true;
+
+		// Clean up after 1 second
+		// TTS triggers copy event immediately
+		setTimeout(function () {
```
keirog: Is there a benefit to setting a timeout here, rather than removing the `return`?

keirog: Ok I think I haven’t understood what’s going on. Bear with me 😄

keirog:
I’ve run this branch locally (in Chrome) but am finding that it lets me copy the article, with no legal notice. Then I tried the master branch locally and that caused a JS error. Are you seeing that on your machine too?

lc512k:
Yeah, there's two issues.

My previous PR was missing the argument in onCopy, which breaks in prod. The fix for that is in this branch. BUT when I try the branch locally the JS breaks because of https issues (and I can't run the router locally for some reason)

What I did to test this worked (🙈) was to copy the JS in the console in prod and run it there.

I'll merge this now just to get the fix in master and if anything's weird I'll just roll back! I don't have it in me to investigate why the router is acting up for me on a friday afternoon 😄

PR #1851, opened 2017-02-13
Fail getInteractive() gracefully (by ignoring it)

adambraimbridge:
Fixes :crossed_fingers: a problem when running locally and npm-linked to @financialtimes/n-ui.

However, I couldn't discern what was making `getInteractive()` — specifically, `interactivePoller.getData()` — fail ...

### DIFF
--- a/server/controllers/negotiation.js
+++ b/server/controllers/negotiation.js
@@ -21,9 +21,12 @@ function isArticleVideo ({ webUrl = '' } = {}) {
 }
 
 function getInteractive (contentId) {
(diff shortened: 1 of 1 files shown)


wheresrhys:
I don't know what the problem is, but this is a dangerous fix - I guess that poller is important, and this code will mean any errors in it will get swallowed in prod. 

Broadly speaking, if the problem only manifests itself when using a particular tool/workflow then the place to fix it should be in the tool ideally, and failing that should be targeted very explicitly at a non prod environment

PR #1853, opened 2017-02-13
Remove webpack from package.json

adambraimbridge:
I need an adult to look at this and make sure I'm not breaking the interwebs. @wheresrhys @GlynnPhillips Thanks for your ongoing support

The question is: Will next-article possibly break in some unexpected way in prod, because it needs these webpack/babel dependencies, that didn't break it locally or in CI?


wheresrhys:
The most important thing to check is to run `make build-production` locally and see what happens - do a file size diff between files built with the old setup and files built with the new setup. If they're the same and everything looks ok when you load the page then you're probably good.

I also recommend a trial run on a non-platinum app too (if that's relevant)

adambraimbridge:
image
We're getting different build sizes when webpack is there vs. removed. I'm installing a diff tool to visualise what the actual differences are. Stay tuned.

adambraimbridge:
Sigh. `master` was commits ahead of my `remove-webpack` branch. I was comparing more than just the absence of webpack in package.json. 

So I merged `master` into `remove-webpack`, and it's looking much better now.

image
^ much closer folder sizes

image
^ Some changes to quotes and an additional comment. Looks fine.

image
^ only `main-without-n-ui.js`. its `.map` equivalent and `__about.json` differ, and the latter is just a timestamp.

adambraimbridge:
That's about `make build`. What about `make build-production`?

image
^ `__about.json`, `asset-hashes.json` and the `.map` files are expected to differ.

wheresrhys:
check it splits up the css files in the same way it does now. That's the most brittle and un-tested bit

adambraimbridge:
I'm doing the same thing in a non-platinum app: https://github.com/Financial-Times/next-feedback/pull/68/files <-- if that goes well, it's good evidence that this also will go well.

adambraimbridge:
- [x] Merged the latest version of `master` into `remove-webpack`
- [x] Repeated `make clear install build` tests
- [x] Saw that the build passes in CI

I'm going to merge.

PR #1864, opened 2017-02-16
WIP: build the article using Next API

adgad:
- [x] Render (more or less) the main article content with GraphQL
- [x] Combine with the read next/suggested reads query
- [x] Render more-ons from server
- [x] Dedupe the more-ons
- [ ] Add the missing fields to Next API (topper, originatingParty)
- [ ] Test all the things (videos, podcasts, light sign up, barriers etc etc etc)


adgad:
So, was testing this locally and requesting a bunch of articles. For the most part it was okay, but graphQL did get a couple of timeouts. Looking at Splunk, yesterday we got 437 graphQL timeouts and <20 ES timeouts, do don't think it's a good idea to move _everything_ to graphQL.

What I will do is:
1) make graphQL onward journey + ES request in parallel
2) When doing story package, compare adding to graphQL vs just doing sequential ES gets vs ES search
3) Keep the bottom more-ons client side and look at lazy loading them.

PR #1866, opened 2017-02-17
set fastft canonical url to next

wheresrhys:
@i-like-robots @sjwells 

I presume this is uncontroversial. Prevents old urls showing up in google and avoids a slow redirect


sjwells:
I don't see any issue with this from the UPP point of view.

PR #1868, opened 2017-02-17
Revert "Revert "set fastft canonical url to next""

wheresrhys:
Reverts Financial-Times/next-article#1867


wheresrhys:
todo
- investigate what happens for old fastft articles not on next yet
- why does google webmaster tools point at the old /fastft site?

PR #1894, opened 2017-03-01
share box fix

gvonkoss:
@adgad @roland-vachter 
This changes the `x` button (from white on a white background to grey on a white background) on the share-box for the comments — I can't seem to find a reference for what it is _meant_ to look like, so please let me know if this change is wrong. 

screen shot 2017-03-01 at 15 56 21

(are there meant to be links/buttons for twitter/facebook in this box?)

### DIFF
--- a/client/components/comments/main.scss
+++ b/client/components/comments/main.scss
@@ -6,10 +6,8 @@ body .fyre-modal .fyre-modal-title-close {
 	background-image: none;
 
 	&:before {
-		@include oIconsGetIcon('cross', #ffffff, 10, $iconset-version: 1);
+		@include oIconsGetIcon('cross', #505050, 20, $iconset-version: 1);
 		content: '';
-		position: relative;
-		top: 1px;
 	}
 }
 
--- a/client/components/comments/main.scss
+++ b/client/components/comments/main.scss
@@ -6,7 +6,7 @@ body .fyre-modal .fyre-modal-title-close {
 	background-image: none;
 
 	&:before {
-		@include oIconsGetIcon('cross', #505050, 20, $iconset-version: 1);
+		@include oIconsGetIcon('cross', oColorsGetPaletteColor('grey-tint4'), 20, $iconset-version: 1);
 		content: '';
 	}
 }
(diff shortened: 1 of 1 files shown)


client/components/comments/main.scss
```diff
@@ -6,10 +6,8 @@ body .fyre-modal .fyre-modal-title-close {
 	background-image: none;
 
 	&:before {
-		@include oIconsGetIcon('cross', #ffffff, 10, $iconset-version: 1);
+		@include oIconsGetIcon('cross', #505050, 20, $iconset-version: 1);
```
onishiweb: if you use `oColorsGetPaletteColor('grey-tint4')` here instead of the hex value it'll make transitioning to the new palette easier in the near future.

client/components/comments/main.scss
```diff
@@ -6,7 +6,7 @@ body .fyre-modal .fyre-modal-title-close {
 	background-image: none;
 
 	&:before {
-		@include oIconsGetIcon('cross', #505050, 20, $iconset-version: 1);
+		@include oIconsGetIcon('cross', oColorsGetPaletteColor('grey-tint4'), 20, $iconset-version: 1);
```
onishiweb: POW! Amazeballs :D

roland-vachter:
Normally, you shouldn't touch it at all :)
I don't see why should comments related styles be overwritten at all.
An example of comments without any overrides can be seen here:
https://ftalphaville.ft.com/2017/03/01/2185313/when-life-imitates-art-uber-ceo-edition/

image

Can you try to remove the overrides completely?

gvonkoss:
I found the overrides! Does that look about right?

client/components/comments/main.scss
```diff
@@ -22,14 +11,14 @@ body .fyre-modal .fyre-modal-title-close {
 
 
 body .fyre-modal .fyre-modal-facebook-icon {
```
roland-vachter: Please remove these ones as well.
Facebook and twitter icons are no longer visible.

PR #1895, opened 2017-03-01
Initial styles for series article - highlight theme, mobile

TheoLeanse:
@adgad @gvonkoss @kavanagh

### DIFF
--- a/client/components/package/main.scss
+++ b/client/components/package/main.scss
@@ -11,8 +11,8 @@ $pkgBlack: #252830;
 
 	margin-top: 2.5em;
 	margin-bottom: 2em;
-	border-top: 1px rgba(0,0,0,0.20) solid;
-	border-left: 1px rgba(0,0,0,0.20) solid;
+	border-top: 1px rgba(0, 0, 0, 0.2) solid;
+	border-left: 1px rgba(0, 0, 0, 0.2) solid;
 	padding: 0.5em 0 0.5em 0.5em;
 	display: block;
 }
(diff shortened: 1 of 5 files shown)


client/components/package/main.scss
```diff
@@ -1,43 +1,104 @@
-.package--nav {
+$pkgYellow: #ffec1a;
+$pkgBlack: #252830;
+
+.package__nav--sticky {
 	position: sticky;
 	top: 56px;
 	z-index: 103 !important;
```
kavanagh: 🤔 why not `102`?

client/components/package/main.scss
```diff
@@ -1,43 +1,104 @@
-.package--nav {
+$pkgYellow: #ffec1a;
```
kavanagh: Will these be in the new o-colors version?

client/components/package/main.scss
TheoLeanse: good question. I was just directly copying the `z-index` of the sticky FT header, but there's some magic going on with that (it's seemingly set to `99` [here][https://github.com/Financial-Times/o-header/blob/704e7a1b5288755081c31d4135eb7cd777e0e462/src/scss/_variables.scss#L24] but becomes `103!important` somewhere between there and my browser...) I will put it back to `99`, unimportant.

client/components/package/main.scss
```diff
@@ -1,43 +1,104 @@
...
+$pkgBlack: #252830;
+
+.package__nav--sticky {
 	position: sticky;
```
adgad: This won't work in older browsers (personally I think that's fine as long as it looks okay - especially as the alternative is JS), but worth checking at some point.

TheoLeanse: yep, needs to be looked into once we are fully settled on the UX of this nav bar (Nick is thinking that it might be good to have it visible only on scroll-up)

client/components/package/main.scss
TheoLeanse: yep!

PR #1897, opened 2017-03-02
Remove experimental content switching

kavanagh:
@adgad @TheoLeanse - is there much point doing this anymore?

It feels like a direct call to CAPI is brittle and using S3 from the article app doesn't satisfy our needs entirely as we also need to mock next-api and next-es-interface behaviour. ie we need the aggregating data provided by Next-API and body transforming and asset fetching of provided the content in Elastic search.


TheoLeanse:
nope, kill it

PR #1898, opened 2017-03-02
Slate offset topper for series articles

adgad:
Jig around some things with the offset topper so it can take a background, and then give it a slate background if it's a series article.

@TheoLeanse @gvonkoss @kavanagh


kavanagh:
Might as well go the whole hog: change `.article-topper` to `.topper`.

adgad:
@gvonkoss pointed out this is all wrong (in that it doesn't follow the design) so need to rethink (╯°□°）╯︵ ┻━┻

PR #1899, opened 2017-03-02
set as canonical url for blogs

wheresrhys:
@richard-still-ft 

Also, I presume we can remove the no robots header now @i-like-robots ?


i-like-robots:
Yep yep

PR #1902, opened 2017-03-02
Lazy load moreons

adgad:
In a hopeful attempt to reduce the number of article 500s we get (half of which come from story packages), I want to try lazy loading the moreons. Doing this behind a flag, using the same code as was used on comments.

@GlynnPhillips @wheresrhys @kavanagh @ironsidevsquincy

### DIFF
--- a/client/components/comments/index.js
+++ b/client/components/comments/index.js
@@ -1,44 +1,9 @@
 import tracking from './tracking';
+import lazyLoad from '../lazy-load';
 
 const oGrid = require('o-grid');
 const commentsIcon = require('./icon');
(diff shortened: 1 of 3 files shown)


kavanagh:
We really need canary releases...

wheresrhys:
2 bits of feedback
- if you're not doing so already, consider varying the thresholds for mobile vs desktop
- have you noticed how sometimes/often images load very slowly. A few weeks ago aI started to dig into it and it seemed, at first glance, like the sending of the request was the thing being slow, not the response time. So are we using intersection observer properly?

PR #1903, opened 2017-03-02
Series article, highlight theme, desktop

TheoLeanse:


client/components/package/main.scss
```diff
@@ -1,22 +1,23 @@
...
 	margin-bottom: 2em;
 	border-top: 1px rgba(0, 0, 0, 0.2) solid;
 	border-left: 1px rgba(0, 0, 0, 0.2) solid;
 	padding: 0.5em 0 0.5em 0.5em;
-	display: block;
 }
+
 .package__nav--article-end {
+	@include oTypographySans(m);
+	background-color: $pkgBlack;
```
adgad: is this the same as `slate` which is in o-colors now?

TheoLeanse: aye, 'tis

client/components/package/main.scss
```diff
@@ -1,113 +1,58 @@
 $pkgYellow: #ffec1a;
-$pkgBlack: #252830;
```
adgad: We should perhaps be smarter about which of these we include in the head - can probably exclude in-article and end-article from there

adgad: (Can do this separately)

client/components/package/contents.scss
```diff
@@ -0,0 +1,39 @@
+$pkgYellow: #ffec1a;
+
+.package__contents {
+	width: 100%;
+	max-width: 490px;
+	position: absolute;
+	right: 0;
+	z-index: 99;
```
adgad: Is there a n-ui z-index helper thing?

adgad: https://github.com/Financial-Times/n-ui-foundations/blob/3eddc639db4e702ca987c6b38ca936c4f9b1dcc9/util/_mixins.scss 

This should probably go in there

PR #1905, opened 2017-03-03
update url

gvonkoss:
@roland-vachter is this along the lines of what you mean, in regards to: https://jira.ft.com/browse/NFT-758?


roland-vachter:
Yep, that looks better :)

PR #1914, opened 2017-03-07
Add themes for in-article series navigation

TheoLeanse:


views/partials/package/package-contents.html
```diff
@@ -2,20 +2,21 @@
 	{{#each this}}
 		{{#ifEquals id @root.context.current.id}}
 			<li class="package__content-item package__content-item--current package__text--m">
-				{{#ifEquals @root.package.sequence 'ordered'}}{{sequenceId}}{{/ifEquals}}
+				{{#if sequenceId}}<div>{{sequenceId}}</div>{{/if}}
```
gvonkoss: does this refer to 
```
tableOfContents {
  "sequence": "..."
}
```
in the model?

TheoLeanse: yep, but it's not coming through yet

gvonkoss: indeedy, the only article that has that is the lorem ipsum one — however, my point was more that it's not called `sequenceID`, it's just `sequence`

TheoLeanse: aha yep I'm using the sequence and the label (is it label or labelType?) to set the sequenceId, which will be like 'PART 1' etc

gvonkoss: `labelType` https://gist.github.com/Nesze/fa4d7672178b24f2f3b8dff4beab3a9d#file-mvp-changes-json

TheoLeanse: does that mean we need to change this? https://github.com/Financial-Times/next-api/blob/6693e0edf354bd479e9ecc0693e9cc811504017e/server/graphql/types/content.js#L142

TheoLeanse: if yes, I'd need to change https://github.com/Financial-Times/next-article/blob/theo/package-article-themes/server/graphql-queries/read-next.js#L57

server/controllers/article-helpers/content-package.js
```diff
@@ -1,5 +1,5 @@
-const placeFirst = ({ first, pkg }) => {
-	return [].concat(first, pkg.filter(x => x !== first));
+const placeFirst = ({ first, all }) => {
+	return [].concat(first, all.filter(x => x !== first));
 };
```
gvonkoss: `x`?

PR #1916, opened 2017-03-08
Topper template update

gvonkoss:


server/controllers/article-helpers/topper-theme-map.js
```diff
@@ -12,47 +12,54 @@ const themeImageRatio = {
...
-			backgroundColour,
-			myFtButtonVariant: myFtButtonVariant(backgroundColour),
-			includesImage: content.topper.theme !== 'full-bleed-text'
-		};
+	let themeOrLayout;
+	let topperOrLeadImages;
+
+	if(content.topper) {
+		//TODO: change to layout only when migration is complete
+		themeOrLayout = content.topper.theme || content.topper.layout;
```
adgad: perhaps these should be the other way round i.e. prefer the new fields to the old ones

PR #1917, opened 2017-03-09
Series topper srsly

adgad:
* Offset stuff all lives in the offset template (not in the body)
* style to allow for backgrounds


gvonkoss:
LGTM

PR #1920, opened 2017-03-09
Ads - Re-instating third-ad on bus-ed section articles

andrewgeorgiou1981:
@adgad


views/partials/bottom/comments.html
```diff
@@ -1,9 +1,14 @@
 <div class="o-grid-row">
-
 	{{#if @root.flags.threeAdProposition}}
 	<div data-o-grid-colspan="12 Lfalse">
 		{{>ads/bottom-mobile}}
 	</div>
+	 {{else}}
+		 {{#if @root.isBusEd}}
```
adgad: I think theres a #ifSome handlebars helper you can use here e.g. {{#ifSome @root.flags.threeAdProposition isBusEd}}

server/controllers/article.js
```diff
@@ -116,7 +116,7 @@ module.exports = function articleV3Controller (req, res, next, content) {
 	// Decorate article with primary tags and tags for display
 	decorateMetadataHelper(content);
 	content.isSpecialReport = content.primaryTag && content.primaryTag.taxonomy === 'specialReports';
-
+	content.isBusEd = (typeof content.metadata.find(tag => tag.idV1 === 'MTI2-U2VjdGlvbnM=') !== 'undefined');
```
adgad: Could we maybe call this isBusinessEducation so it's a bit clearer?

PR #1927, opened 2017-03-10
Labelling

adgad:
@TheoLeanse @kavanagh

### DIFF
--- a/server/controllers/article-helpers/content-package.js
+++ b/server/controllers/article-helpers/content-package.js
@@ -4,10 +4,10 @@ const placeFirst = ({ first, all }) => {
 	return [].concat(first, all.filter(x => x !== first));
 };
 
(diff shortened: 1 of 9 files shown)


server/controllers/article-helpers/content-package.js
```diff
@@ -37,7 +44,8 @@ module.exports = ({ id, containedIn }) => {
 	const pkg = containedIn[0];
 	const currentIndex = pkg.contains.findIndex(item => item.id === id);
 	const contents = addContents({ pkg, currentIndex });
-	const context = addContext({ pkg, currentIndex });
+
+	const context = addContext({ pkg, currentIndex, contents });
```
TheoLeanse: so we're not using `contents` now riiight?

adgad: Indeed, removed

PR #1931, opened 2017-03-10
/in-article-barrier endpoint for injecting barriers from `next-product`

remybach:


tavvy:
Great, could we also remove use of the `ft-access-preview` header and the code associated with it. Let me did up a few PRs for reference ...

PR #1936, opened 2017-03-13
Remove the special reports related links

adgad:
@TheoLeanse @gvonkoss @kavanagh 

screen shot 2017-03-13 at 12 06 41

Removing this from the bottom of all special report articles. They will now fall back to use the normal more-ons.

Usage was also quite low - 1987 users clicking them in the last 14 days.


gvonkoss:
SUCH DELETES 👍

PR #1940, opened 2017-03-13
Delete tearsheets and moar

adgad:
* Tearsheet hovers - flag expired 144 days ago
* Reading history - stored UUIDs in session storage, was used for an old myFT feature back when I was in that team
* Suggested reads (RHR) CSS no longer needed as they are just teasers now 
* Cache journey (caching onward journey in SW) - code commented out so deleting
* Ad blocking controller from the labs prototype, but not referenced in code (/cc @seanmtracey )
* We had some code to style o-share buttons, but that is now in o-share
* onward journey header - not used anywhere, just using teaser collection stuff now
* article-pod-mapping-v3 - no idea what this is but not used anywhere


views/partials/right-hand/right-hand.html
```diff
@@ -4,7 +4,6 @@
 
 	{{#ifAll package @root.flags.contentPackages}}
 	{{else}}
-		{{{tocHTML}}}
```
ironsidevsquincy: no more TOC..?

adgad: Nope, was hardly being used, gets confusing with content package navigation and we're probably going to re-look at them properly later anyway.

adgad: (deleted the rest of the code in an earlier PR)

client/components/tearsheets/intent.js
```diff
@@ -1,16 +0,0 @@
...
-	target.addEventListener('mouseenter', () => {
-		clearTimeout(timeout);
-		timeout = setTimeout(callbackIn, INTENT_ENTER);
-	});
-
-	target.addEventListener('mouseleave', () => {
-		clearTimeout(timeout);
-		timeout = setTimeout(callbackOut, INTENT_LEAVE);
-	});
-}
```
i-like-robots: aww, this code never stood a chance

server/transforms/extract-toc.js
```diff
@@ -1,11 +0,0 @@
...
-
-module.exports = function (bodyHTML) {
-
-	const $ = cheerio.load(bodyHTML);
-
-	return {
-		tocHTML: $.html($('.table-of-contents').remove()),
-		bodyHTML: $.html()
-	};
-};
```
i-like-robots: This must be some of the oldest next code

server/mappings/ad-blocking-articles-template.js
```diff
@@ -1,13 +0,0 @@
-module.exports = article => {
```
seanmtracey: Sorry about the dead code. Thought I'd removed all of the ad-blocking stuff long ago.

PR #1948, opened 2017-03-14
use a package's brand where available

TheoLeanse:
@adgad I think this should just work?? 😬


adgad:
Isn't primaryBrandTag an object? i.e. should it be primaryBrandTag.prefLabel or something like that?

adgad:
Saying that if it works then 👍

TheoLeanse:
@adgad closing this, cos it's done?

PR #1961, opened 2017-03-17
Tidy up some css

TheoLeanse:
@gvonkoss


client/components/package/head.scss
```diff
@@ -1,32 +1,4 @@
...
-	.package__highlight-border.package__highlight-border,
-	.package__article-end-teaser .package__highlight-border.package__highlight-border:after {
-		border-color: $pkgYellow;
-	}
-
-	.o-teaser__heading {
-		color: getColor('white');
-	}
-}
+@import 'colors';
```
gvonkoss: if you import colors before everything else, I think you might not need it in the other files?

TheoLeanse:
```
.package__expander {


}

.package__expander[aria-expanded="true"] {

}

PR #1962, opened 2017-03-20
Package teasers

gvonkoss:


client/components/package/head.scss
```diff
@@ -1,4 +1,5 @@
+@import 'colors';
+@import 'teasers';
```
TheoLeanse: this `head.scss` is just importing the stuff necessary to load the top nav, so I think we could get away with importing the teasers styles into the `main.scss` as they're only used in the bottom navigation

views/partials/teasers/end-nav.html
```diff
@@ -0,0 +1,23 @@
...
+			</a>
+			<p class="o-teaser__standfirst">{{standfirst}}</p>
+		</div>
+		<div class="package__end-title-container">
+			<span class="package__end-nav-title package__text--m package__text--strong">Explore the series</span>
+		</div>
+		<div class="o-teaser__image-container">
+			<div class="o-teaser__image-placeholder image-placeholder">
+				<a href="{{relativeUrl}}">
+					<img src="{{mainImage.url}}" class="o-teaser__image" alt="demo image"></img>
```
TheoLeanse: wonder if we're going to need to pass a value for the alt text into this teaser...

client/components/package/_teasers.scss
```diff
@@ -0,0 +1,60 @@
...
+		.o-teaser {
+
+			&__meta,
+			&__duration,
+			&__heading {
+				color: oColorsGetPaletteColor('white');
+			}
+
+			&__standfirst {
+				color: oTeaserTextColor(oColorsGetPaletteColor('white'), oColorsGetPaletteColor('slate'), 70);
```
TheoLeanse: cool, what does this do?

PR #1963, opened 2017-03-20
Use signed fetch

i-like-robots:
Drops use of [next-ft-api-client](https://github.com/Financial-Times/next-ft-api-client) as it has long been deprecated.

Also updates all the fixtures affected tests along with it which makes this PR kinda big.


server/controllers/content.js
```diff
@@ -13,14 +13,18 @@ function getInteractive (contentId) {
...
-		index: 'v3_api_v2'
-	})
-	// Some things aren't in CAPI v3 (e.g. Syndicated content)
-		.catch(function (error) {
+	return signedFetch(`https://next-elastic.ft.com/v3_api_v2/item/${contentId}`)
+		.then((response) => {
+			if (response.ok) {
+				return fetchres.json(response);
+			} else {
+				throw new fetchres.BadServerResponseError(response.statusCode);
```
kavanagh: So all 4xx response would throw? That means we'll never get to see if Falcon can render the article.

Either we bite the bullet and so redirecting to Falcon or we move the 'see if Falcon has it' logic to the the catch all for most article errors.

i-like-robots: The logic here should be the same as previous - the lib I'm swapping out uses fetchres in the same way.

The spec for this still passes and a quick test shows we're still good:

```sh
$ curl -I https://localhost.ft.com:5050/content/8f88c930-d00a-11da-80fb-0000779e2340
HTTP/1.1 302 Found
X-Host: http://localhost:3002
location: http://www.ft.com/cms/s/0/8f88c930-d00a-11da-80fb-0000779e2340.html?ft_site=falcon
```

i-like-robots:
Build fails to to pA11y errors

PR #1967, opened 2017-03-21
Little ux fixes for Package nav in articles

TheoLeanse:
- [x] maximum length of packages from 6 to 9
- [x] see more parts | stories depending on package ordering
- [x] link to PLP wherever the series title can be seen
- [x] investigate focus on the top nav expander
- [x] top nav dropdown is overlay only on desktop (otherwise it pushes the content down)
- [ ] transitions on expander open
- [x] handle current article differently in menu for unordered packages
- [x] remove drop down menu from the mid article navigation, replace with link to PLP
- [x] line break between brand and package title in middle component
- [x] remove the mid nav


adgad:
👍

PR #1971, opened 2017-03-22
Special report labelling

adgad:
@TheoLeanse 

Sorry if there are conflicts. Might be one or two things that I've left which still have series hardcoded, but can look at that after your changes.


TheoLeanse:
👍

PR #1975, opened 2017-03-23
Degrade package pages when Next API fails

kavanagh:
@adgad @TheoLeanse 

If Next API fails OR the flag is off
* Render PLP _intro_ articles as regular articles but try to retain the package topper.
* Redirect PLP _index_ pages to the first page in the package, if there is one.
* Don't redirect PLP _index_ pages if the are no article pages. We assume this wouldn't really happen except in Preview. If it did appear in prod the normal onward journey would show and probably be helpful(ish).


server/controllers/content.js
```diff
@@ -87,18 +87,29 @@ module.exports = function contentController (req, res, next) {
...
-			// 	}
-			// }
+			if (content.type === 'package') {
 
+				// Allow us to degrade gracefully
+				if (!onwardJourney) {
+					content.degradePackagePage = true;
+				}
+
+				{
```
TheoLeanse: is there a missing statement before these curlies?

PR #1977, opened 2017-03-23
Amend: switch sourcepoint flag off for smoke tests

andygnewman:
Sourcepoint is what we use to identify ad blocking behaviour and uses pixels to test browser response and is expanded in n-ui 5.2.0.
These pixels fail a11y tests.
So this switches the relevant flag off in the a11y tests.


adgad:
Out of curiosity, what do they fail?

andygnewman:
The dummy pixels are img tags and don't have alts, which is a no-no for accessibility. So just like we turn ads flag off, we need to turn this one off aswell.

adgad:
👍  Worth raising with Sourcepoint as well, and trying to get them to fix.

andygnewman:
Their script does pull the pixels out once the check is complete, but the timing conflicts with the a11y test being run.

adgad:
Ahh gotcha :)

PR #1980, opened 2017-03-23
Remove more-ons from package articles

adgad:
@TheoLeanse @gvonkoss


client/main.scss
```diff
@@ -189,6 +189,14 @@ $o-video-is-silent: false;
 	}
 }
 
+//the gridded teaser-collection used on packages need extra spacing between them
+.o-teaser-collection--square {
```
kavanagh: It confused me that this is here and not in o-teaser-collection. Is the plan to add this to o-teaser collection?

adgad: Possibly, at some stage. It's a little weird because the teasers at the moment have a 15px margin-bottom - i'm not sure yet if this extra spacing is a one off for PLPs.

Maybe i'll not use the `o-teaser-collection--square` class and call it something specific for this case.

PR #1982, opened 2017-03-24
Begin non-special-report landing pages.

kavanagh:
This adds article copy to the PLP and the new layout for the teasers. 

The slate teaser styles are here just to get started with but I'll move them in n-teaser next.


client/components/article/_main.scss
```diff
@@ -30,6 +30,23 @@
 
 .article__main {
 	padding-top: 15px;
+	padding-bottom: 30px;
+	&-content {
+		@include oGridCenter;
```
adgad: Should these be __ ?

server/model/package.js
```diff
@@ -83,13 +100,11 @@ module.exports = function decoratePackage (req, res, payload, flags) {
 					case 'special-report':
```
adgad: Is this redundant now (there's something similar above)?

kavanagh: The messages are slightly different depending on the teaser collection.

client/components/article/_main.scss
kavanagh: I'll change it

PR #1985, opened 2017-03-24
Add logic to detect if article model is for truncated article

tavvy:
and hide the light signup component if it is true

- add test coverage

----

Previously we would show the lsu component AND the barrier component.


adgad:
:+1 does it need to go on the other types of content pages (video, content package, podcasts etc)

tavvy:
nope, just articles 👍

PR #1986, opened 2017-03-27
Remove unnecessary package theme code

TheoLeanse:


client/components/package/_top-nav.scss
```diff
@@ -1,13 +1,14 @@
-@import 'expander-themes';
+$navbar-height: 40px;
+$desktop-gutter-width: 16px;
```
adgad: I think there's an o-grid helper for this...

https://github.com/Financial-Times/o-grid#gutters

PR #1988, opened 2017-03-27
Handle an article not actually contained in its package

TheoLeanse:
@adgad @kavanagh not sure if this is really needed...


TheoLeanse:
issue is 
screen shot 2017-03-27 at 10 28 19

PR #2008, opened 2017-03-30
➕ Adds better tracking to benchmark survey; shows only to logged in users

aendrew:


client/components/benchmark-survey/index.js
```diff
@@ -25,8 +26,8 @@ export default function benchmarkSurvey (flags, opts = {
 
 	let shownCount = Number(window.localStorage.getItem(`${opts.sideboxId}-show-count`)) || 0;
 	const isRemoved = window.localStorage.getItem(`${opts.sideboxId}-ad-removed`) || false;
-
-	if (hasLocalStorage && !isRemoved && shownCount <= opts.maxShowCount ) {
+	const isLoggedInUser = !!~document.cookie.indexOf('FT_User'); // Show only if authenticated user
```
adgad: FTSession is probably a better cookie to check for here (FT_User I believe is deprecated).

myFT do something like this - `!!(/FTSession=/.test(document.cookie))`

ironsidevsquincy: You can also use the [next-session-client](https://github.com/Financial-Times/next-session-client#usage) to get the session Id (get's the secure version)

ironsidevsquincy: Or get the uuid if you wan't to be absolutely certain they're signed in

PR #2009, opened 2017-03-30
Fix: Social share buttons that follow article content

andygout:
Fixes: [WhatsApp button at bottom of article is broken](https://jira.ft.com/browse/NFT-792)

Also means that behaviour between both sets of buttons is consistent, e.g. currently top Twitter link opens new window whereas the bottom opens new tab; fix will apply `o-share` logic to both and make bottom link also open in new window (not tab).


client/components/share/main.js
```diff
@@ -4,9 +4,17 @@ import { EmailArticleData, emailArticleModes, EmailArticleView } from '@financia
...
+
+	for (let i = 0; i < shareContainers.length; i++) {
+
+		if (shareContainers[i] && !shareContainers[i].classList.contains('data-o-share--js')) {
+
+			new OShare(shareContainers[i]);
+
+		}
+
 	}
```
bjfletcher: aha, good find... nice, nice

another way of doing that:

```js
document.querySelectorAll('[data-o-component=o-share]')
  .filter(share => share.classList.contains('data-o-share--js'))
  .forEach(share => new OShare(share));```

either is grand :)

andygout: `document.querySelectorAll(...).filter is not a function`

Result of `querySelectorAll` is going to be a NodeList not an Array, so cannot call Array-exclusive methods (`filter` and `forEach`) on it.

ironsidevsquincy: There's a `$$` function in n-ui-foundations which calls `querySelectorAll` and returns an actual `Array` - https://github.com/Financial-Times/n-ui-foundations/blob/master/main.js#L11

andygout: Nice - have used that with Ben's solution here: https://github.com/Financial-Times/next-article/pull/2012

PR #2012, opened 2017-03-30
Use: n-ui querySelectorAll shortcut for social share button fix

andygout:


client/components/share/main.js
```diff
@@ -1,21 +1,14 @@
...
-	for (let i = 0; i < shareContainers.length; i++) {
-
-		if (shareContainers[i] && !shareContainers[i].classList.contains('data-o-share--js')) {
-
-			new OShare(shareContainers[i]);
-
-		}
-
-	}
+	$$('[data-o-component=o-share]')
```
ironsidevsquincy: if you were feeling fruity, could have used `:not` here instead of the filter below

andygout: Hmm, think I am feeling fruity. Take 3 coming up.

PR #2027, opened 2017-04-04
Remove in-article-barrier stuff

remybach:


tavvy:
Failures are due to the now missing css files for barriers.
You may need to remove references in places like this too:
https://github.com/Financial-Times/next-article/blob/master/.gitignore#L17

